PR #89, opened 2016-01-06
Fixing linking issues when using C99

pavanky:
Fixes issues with gcc 5.x which use c99 as the default C standard.


yzhwang:
Thanks Pavan!

Issue #69, opened 2015-11-30
If you pass a bogus filename as the input matrix, you get a non-helpful output


jowens:
`Unable to open file: No such file or directory
Unable to open file: No such file or directory
[/home/jowens/working/gunrock/gunrock/util/array_utils.cuh, 510 @ gpu 0] row_offsets cudaMemcpy H2D failed (CUDA error 11: invalid argument)`

A much more meaningful error message would be nice. Perhaps assign to one of the new students?

wetliu:
I want to try this problem. How can do I need to get assigned?

jowens:
@wetliu On the right side of this page is "Assignee". Assign yourself!

sgpyc:
One more thing about input: if passing invalid arguments, like -undirected (missing one -), --undirect (wrong argument), --undirected --undirected (twice), --device=0, --device=1,2 (twice with different values), the behavior now is not as expected.

jowens:
@sgpyc that should be a different issue, don't wedge it into this one

sgpyc:
Moved to issue #91 , thanks @jowens

wetliu:
The gunrock has the exception case for inputting the wrong file name. Thanks for the help from gunrock team, I edited a little bit and right now it can catch the bogus file name. 

However, I need some help in identifying what the main function will do if the LoadGraph is has an error. The output right now is this:

weitang@slaking:~/gunrock/tests/bc/bin$ ./test_bc_7.5_x86_64 market ../test_bc.mLoading Matrix-market coordinate-formatted graph ...
Input graph does not exist.
Using 1 GPU: [ 0 ].
 [../../gunrock/util/array_utils.cuh, 510 @ gpu 0] row_offsets cudaMemcpy H2D failed (CUDA error 11: invalid argument)
Computing reference value ...
Segmentation fault (core dumped)

I hope to stop executing the function afterwards instead of getting the seg fault at the end. Thank you!

sgpyc:
Maybe capture the cudaError_t returned by info->Init("BFS", *args, csr);, if it's not cudaSuccess, then stop the program / throw an exception. Same checking should be added to Init and Reset for both problem and enactor.

sgpyc:
Invalid filename now stops the program, with a suitable error message.

```
sgpyc@mario:~/Projects/gunrock_master/examples/bfs$ ./bin/test_bfs_9.1_x86_64 market /data/haha
Loading Matrix-market coordinate-formatted graph ...
[...]

Issue #85, opened 2015-12-30
Codify best practices wrt git and branches for Gunrock development


jowens:
My impression on our best-practice was:
- We are trying to follow this: http://nvie.com/posts/a-successful-git-branching-model/
- Master should correspond to actual releases, i.e., what's checked into master should have a version x.y.z. It should of course be stable and pass ctest. It should give us stable but good performance.
- Dev should also be stable and pass ctest. But it might have features in it that haven't been integrated into master yet. Generally we work within dev.
- If you're making changes that will break something, then make your own branch, work in it until it passes ctest, then merge it into dev.
- We use semantic versioning. http://semver.org/

neoblizz:
We need to documents this and some other things for gunrock developers.

neoblizz:
New method suggested, let's move the discussion there: https://github.com/gunrock/gunrock/pull/267

PR #94, opened 2016-01-24
This is for gunrock issues #69

wetliu:
Has changed the if branch for the bogus file name and added the exit part. Not sure yet if exit is the best because a lot of people say that exit is not as good as return, but using return has to change main().

### DIFF
--- a/gunrock/app/enactor_base.cuh
+++ b/gunrock/app/enactor_base.cuh
@@ -646,11 +646,14 @@ public:
             {
                 printf("Loading Matrix-market coordinate-formatted graph ...\n");
             }
+
             char *market_filename = args.GetCmdLineArgvDataset();
-            if (market_filename == NULL)
+
+            std::ifstream fp(market_filename);
+            if (market_filename == NULL||!fp.is_open())
             {
                 fprintf(stderr, "Input graph does not exist.\n");
-                return 1;
+                exit (EXIT_FAILURE);
             }
             boost::filesystem::path market_filename_path(market_filename);
             file_stem = market_filename_path.stem().string();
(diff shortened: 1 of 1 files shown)


jowens:
@yzhwang @wetliu Shouldn't this be a merge into dev not master?

gunrock/app/enactor_base.cuh
```diff
@@ -646,11 +646,14 @@ public:
...
                 printf("Loading Matrix-market coordinate-formatted graph ...\n");
             }
+
             char *market_filename = args.GetCmdLineArgvDataset();
-            if (market_filename == NULL)
+
+            std::ifstream fp(market_filename);
+            if (market_filename == NULL||!fp.is_open())
             {
                 fprintf(stderr, "Input graph does not exist.\n");
```
jowens: I would print the filename here too.

yzhwang: I agree with @jowens on adding the filename to stderr. Have you run ctest after the change? Also, please test for all primitives for a bogus filename or an empty filename, and see if the output make sense. (You can post the output here, you can and you should write a simple shell/python script for your test too)

As for pushing to the master, @wetliu I might have told you to push to the master, if so I apologize, according to https://github.com/gunrock/gunrock/issues/85 master is for actual release. So please cherrypick this commit to the dev branch of your fork and do the things I suggested, then submit the pull request again. Thank you!

wetliu:
I am sorry that it's my mistake to merge it to master. I can print a file name there. I have not yet written a shell script. I will do it later. Thank you for your comments @jowens @yzhwang

Issue #76, opened 2015-12-16
For non-ubuntu linux distributions, handle METIS dependency properly...


yzhwang:
...by either adding code in CMakeLists.txt for users to input their own METIS installation directory, or providing a InstallMetis script.

neoblizz:
Solved with the latest merge, I will test the scripts on more OS just to see how it handles (maybe on OSX too).

PR #98, opened 2016-01-28
Addresses issue #76, allows users to install metis or boost.

neoblizz:
CMakeLists.txt now gives a warning if metis was not found, and also directs the users to the `dep/` directory where shell scripts for installing metis or boost on non-ubuntu linux users can be found.


neoblizz:
@yzhwang Fixed the CMakeLists.txt, now the search for Metis library is working.
@sgpyc  Why don't we include metis library dirs when it is found (like we do for boost or openmp)?

PR #115, opened 2016-02-11
Update CMakeLists required Boost version for the dev branch

ptheywood:
CMakeLists currently requires boost 1.53, which does not include `boost/predefs.h` used in `gunrock/app/enactor_base.h`.

This requires boost 1.55+, but seeing as #114 set it to 1.58 I did the same.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ set(gunrock_VERSION_MINOR 3)
 set(gunrock_VERSION_PATCH 0)
 add_definitions("-DGUNROCKVERSION=${gunrock_VERSION_MAJOR}.${gunrock_VERSION_MINOR}.${gunrock_VERSION_PATCH}")
 
-set(gunrock_REQUIRED_BOOST_VERSION 1.53)
+set(gunrock_REQUIRED_BOOST_VERSION 1.58)
 set(gunrock_REQUIRED_METIS_VERSION 5.0)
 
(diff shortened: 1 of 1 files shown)


neoblizz:
@yzhwang Maybe we should have the lowest possible as required? Which I believe is 1.55 right now, but I am not sure if we will eventually end up using 1.58 ( @sgpyc )?

yzhwang:
@sgpyc please confirm this, if our current enactor_base implementation requires a minimum version of boost 1.58, we should update the documentation and CMakeLists in our master branch too.

sgpyc:
The main thing to consider is, which version of boost is compatible with nvcc 7.5 and c++11. I don't know we have used anything specific in boost that requires high version number.

neoblizz:
Pretty sure 1.54 isn't anymore, @wetliu is using 1.55 I believe and it works for him.

yzhwang:
If for dev it is the case, I agree that we should make it the lowest possible. Please confirm that 1.55 is the minimum requirement and either @neoblizz or @ptheywood could submit the PR if you'd like to. Thank you guys!

neoblizz:
Correction, we found out that the dev version runs on 1.56. I will just run a test to make sure that is the case and then submit a pull request with updated required CUDA version (which now is 7.5).

neoblizz:
Fixed with pull request #130, Boost updated to 1.56.

Issue #97, opened 2016-01-28
Metis is required right now for Gunrock installation but it shouldn't be...


neoblizz:
`find_package(Metis 5.0 REQUIRED)` this doesn't give an error/warning in CMakeLists, I will have it corrected in issue #76 where it will warn the users. 

But after talking to @sgpyc it is not suppose to be required, Gunrock should be able to work as long as it doesn't use Metis. Support can be added where Metis `includes` can be commented for use without this dependency.

jowens:
oops

jowens:
I support fixing this.

neoblizz:
@sgpyc (or @yzhwang) I would like to try this, can you point me to the places where metis is included (or if there is a way I can search that on my own)?

sgpyc:
It's here https://github.com/gunrock/gunrock/blob/dev/gunrock/app/metisp/metis_partitioner.cuh#L17

neoblizz:
Ok, so this is really cool. I worked on this issue @sgpyc and here's my fix for not forcing users to include/use `metis_partitioner.cuh`:
Edits to `CMakeLists.txt` so that it adds a pre-compiler flag `METIS_FOUND` that can be checked in the `metis_partitioner` file using `add_definitions`:

```
FIND_LIBRARY(METIS_LIB NAMES metis)
find_package(Metis ${gunrock_REQUIRED_METIS_VERSION} REQUIRED)
SET(METIS_FOUND FALSE)
IF (METIS_LIB)
  SET(METIS_FOUND TRUE)
  add_definitions( -DMETIS_FOUND=${METIS_FOUND}) #Adds compiler definition METIS_FOUND = TRUE
  message(STATUS "Found Metis")
ELSE (METIS_LIB)
  add_definitions( -DMETIS_FOUND=${METIS_FOUND}) #Adds compiler definition METIS_FOUND = FALSE
  message(WARNING "Metis was requested but support was not found, run /dep/install_metis.sh for installation.")
ENDIF (METIS_LIB)
```

Then, just add `#ifdef METIS_FOUND ... #endif` in the  `metis_partitioner` file. Now, here's the cool thing that I didn't know about! You can actually output pre-compiler definitions using the following code, during `make` it tells me if Metis was found by checking the `METIS_FOUND` and prints `TRUE` or `FALSE`:
```
#define XSTR(x) STR(x)
#define STR(x) #x

#pragma message "Metis Found? " XSTR(METIS_FOUND)
```
[...]

PR #138, opened 2016-02-24
Added support to not use Metis partitioner. Metis is now optional for…

neoblizz:
… gunrock. @sgpyc Can you please show me how to test, or test this yourself? I am not sure how to use `metis_partitioner.cuh` file to invoke the changes I have made. I tried `cmake` and compiling these changes and they work fine! It is suppose to print an error saying you cannot use metis partitioner if you do not have Metis installed. (Issue #97)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,10 +38,10 @@ FIND_PACKAGE(Metis ${gunrock_REQUIRED_METIS_VERSION} REQUIRED)
 SET(METIS_FOUND FALSE)
 if (METIS_LIB)
   set(METIS_FOUND TRUE)
-  add_definitions( -DMETIS_FOUND=TRUE )
+  add_definitions( -DMETIS_FOUND=true )
   message(STATUS "Found Metis")
 else (METIS_LIB)
-  add_definitions( -DMETIS_FOUND=FALSE )
+  add_definitions( -DMETIS_FOUND=false )
   message(WARNING "Metis was requested but support was not found,
   run `sudo apt-get install metis` or `/dep/install_metis.sh` for installation.")
 endif (METIS_LIB)
@@ -325,4 +325,3 @@ set_tests_properties(SHARED_LIB_TEST_PAGERANK
 #    data/simple_example/bips98_606.mtx --src=largestdegree)
 #set_tests_properties(TestBFS
 #    PROPERTIES PASS_REGULAR_EXPRESSION "CORRECT")
-
--- a/gunrock/app/metisp/metis_partitioner.cuh
+++ b/gunrock/app/metisp/metis_partitioner.cuh
@@ -12,6 +12,14 @@
  * @brief linkage to metis partitioner
  */
 
+/* Temporary fix for using Makefiles
+ * in tests/primitive_name/
+ */
+#ifndef METIS_FOUND
+  #define METIS_FOUND false
+#endif
+/* METIS_FOUND default false */
+
 #pragma once
 
 #ifdef METIS_FOUND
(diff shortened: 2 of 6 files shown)


neoblizz:
Also, fixed typos for `--partition-method` flag in all the test files.
Problem with metis partitioner is that I get the flag to enable or disable metis from CMakeLists.txt, but if the user only uses the `tests/bfs` and directly make there, the flag doesn't exist. Any ideas how I can fix that? @yzhwang is looking into `cmake` because we shouldn't have to change so many cmake files because our main file changed.

yzhwang:
@neoblizz I think we should get rid of all the CMakeList.txt under each test/primitive_names directories. I figured when we run cmake once, it also adds Makefile to build_dir/tests/primitive_names (note, not gunrock/tests/ but build_dir you have chosen for all cmake generated files). Under those directories, you can actually run make to build each single primitive. I'm gonna have to talk to Andy and Yuduo whether it makes sense to get rid of all these cmakelist.txt since they are the ones who added them in the summer of 2014. But it seems that now we do have ways to build single primitive and we only need to keep one copy of cmakelist under the root directory.

neoblizz:
@yzhwang Yes, I totally agree. Because making changes in the root `CMakeList.txt` doesn't translate into others, for example the issue I just solved with openmp support on OSX is currently only fixed for the root `CMakeList.txt`, to solve for rest it requires me to change in every single cmake file in the tests dir.

sgpyc:
There is a master Makefile in tests, you can just change that

neoblizz:
@sgpyc Yes, but we give people option to also use cmake using the CMakeLists.txt files, do we really need those?

jowens:
Please make everything work with cmake. Not only is it cross-platform, it's got lots of helpful subroutines (like find openmp, etc.) that make it way easier to make anything complex work more easily.

neoblizz:
I agree, that's how the case should be. I think @yzhwang and I were suggesting that the root CMakeList.txt should create the make files so they stay updated with whatever future changes we make.

jowens:
I concur.

sgpyc:
Does Cmake has the "include" feature? I think that's what may be useful.
And if you want to change or generate the Makefiles, it saves some work by doing it on the master Makefile, rather than AL the individual ones.

yzhwang:
@neoblizz @sgpyc actually I was wrong about removing the CMakelist.txt under tests/primitive_name since they do get included here:
https://github.com/gunrock/gunrock/blob/master/CMakeLists.txt#L211
So now we are using add_subdirectory but since @neoblizz mentioned you have made some changes in the root CMakeLists.txt and they didn't propagate to individual CMakeLists.txt. I think you should make those as functions and put them in .cmake files so that they can be included in any CMakeLists.txt.
here is a post on the difference between add_subdirectory and include.
https://cmake.org/pipermail/cmake/2007-November/017897.html

Seems that we want to put the code that handles dependency issue in a .cmake file so that we can include in both the root CMakeLists.txt and several individual ones. Then, when we want to build only one primitive, we create a primitive_name_build dir, enter it and call:
cmake gunrock_dir/tests/primitive_name/
it will load the CMakeLists.txt under that primitive directory. Now when I do this, it complains several things since we probably don't have the correct CMakeLists.txt defined there. I will do some more experiments tonight.

neoblizz:
I see, that makes sense. Separate the code in root CMakeLists.txt into individual modules that can be included in all CMakeLists.txt files. Question, the [Makefiles](https://github.com/gunrock/gunrock/blob/master/tests/bfs/Makefile) in the tests are generated through the CMakeLists.txt in the test? If not, then how will they get all the dependencies or do we even need them there once we have the CMakeLists.txt updated?

yzhwang:
No, they are written by me before we used cmake. When we finished all the cmake updates, we can get rid of the individual makefiles under each primitive dir.

sgpyc:
But they are useful for debugging purpose. Unless there is a good and easy way to know and change the actual make comment parameter by changing CMakeList.txt, I suggest to keep Makefile, and generate / change them by cmake

Issue #139, opened 2016-02-29
CMakeLists.txt files should update based on root CMakeLists.txt...


neoblizz:
Thinking about separating the CMakeLists.txt stuff into `.cmake` modules and including them in all CMakeLists.txt.

jowens:
Yes.

neoblizz:
Findings; The structure we have right now is actually pretty good. I have been working on separating the modules into .cmake include files, which can easily be included in the root `CMakeLists.txt`. It generates a Makefile for the whole project and Makefiles for the tests primitives in the same build dir. 

Now, here's an issue. @jowens we were hoping you could ask cmake people if there is a better way to implement something like this. What is happening is if we want tests/primitives as separate sub-projects that can be compiled using the individual CMakeLists.txts in their folders we literally find ourselves copy pasting almost the entire root file CMakeLists.txt per primitive (even if it is with the includes). 

What @yzhwang and I were wondering was if there is a better way to have a main project and then sub-projects that get the shared dependencies from main project and can be compiled. Our directory structure is;

CMakeLists.txt (root project)
| __ bc
| __ | __ CMakeLists.txt (sub-project)
| __ bfs
| __ | __ CMakeLists.txt (sub-project)
| __ cc
| __ | __ CMakeLists.txt (sub-project)
...

Right now, if we include almost every necessary library at the end I get this error:
```
[100%] [100%] Building CXX object CMakeFiles/breadth_first_search.dir/home/muhammad/gunrock/externals/moderngpu/src/mgpuutil.cpp.o
Building C object CMakeFiles/breadth_first_search.dir/home/muhammad/gunrock/gunrock/util/gitsha1.c.o
Linking CXX executable breadth_first_search
/usr/bin/ld: CMakeFiles/breadth_first_search.dir/breadth_first_search_generated_test_bfs.cu.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
CMakeFiles/breadth_first_search.dir/build.make:422: recipe for target 'breadth_first_search' failed
make[2]: *** [breadth_first_search] Error 1
CMakeFiles/Makefile2:60: recipe for target 'CMakeFiles/breadth_first_search.dir/all' failed
make[1]: *** [CMakeFiles/breadth_first_search.dir/all] Error 2
Makefile:75: recipe for target 'all' failed
[...]

PR #140, opened 2016-03-01
Improved CMakeLists.txt support for the entire project to...

neoblizz:
include the following changes;
- Separated Metis, Boost, OpenMP, mGPU and CUB into their individual `.cmake` files, now we can just include them wherever we need them. In the root `CMakeLists.txt` they are all included.
- Commented a lot of code in `CMakeLists.txt` to basically explain the important sections if further edits are required in future, I have cleaned up a lot of stuff in there.
- Added the following code to suppress the warnings due to `auto_ptr` in moderngpu:
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-deprecated-declarations")
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
add_definitions ("-Wno-deprecated-declarations")
``` 
- Commented `find_package(CUDA REQUIRED)` in `bfs/CMakeLists.txt` because it is redundant and causes extra message echo in the root cmake compilation.

Everything was tested and worked beautifully, without any annoying warnings from moderngpu. :) This pull request addresses part of issue #139. Later part of compiling sub-projects I am still working on, but that is a different pull request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,113 +5,51 @@ set(gunrock_VERSION_MINOR 3)
 set(gunrock_VERSION_PATCH 0)
 add_definitions("-DGUNROCKVERSION=${gunrock_VERSION_MAJOR}.${gunrock_VERSION_MINOR}.${gunrock_VERSION_PATCH}")
 
-set(gunrock_REQUIRED_CUDA_VERSION 7.5)
-set(gunrock_REQUIRED_BOOST_VERSION 1.56)
-set(gunrock_REQUIRED_METIS_VERSION 5.0)
-
 cmake_minimum_required(VERSION 2.8)
+
+# begin /* Suppress "deprecated auto_ptr" warnings caused by moderngpu */
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-deprecated-declarations")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-declarations")
+add_definitions ("-Wno-deprecated-declarations")
+# end /* Suppress "deprecated auto_ptr" warnings caused by moderngpu */
 
 option(CMAKE_VERBOSE_MAKEFILE ON)
 
+# begin /* Find and set CUDA arch */
+set(gunrock_REQUIRED_CUDA_VERSION 7.5)
 FIND_PACKAGE(CUDA ${gunrock_REQUIRED_CUDA_VERSION} REQUIRED)
-
 if(CUDA_64_BIT_DEVICE_CODE)
   set(gunrock_arch_suffix x86_64)
 else()
   set(gunrock_arch_suffix i386)
 endif()
+# end /* Find and set CUDA arch */
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-
-FIND_PACKAGE(Boost ${gunrock_REQUIRED_BOOST_VERSION}
-  REQUIRED system filesystem timer chrono)
-if (Boost_FOUND)
-  include_directories(${Boost_INCLUDE_DIRS})
-  link_directories(${Boost_LIBRARY_DIRS})
-else()
-  message(WARNING "Boost was requested but support was not found, run
-  `sudo apt-get install libboost1.58-all-dev` `/dep/install_boost.sh` for installation.")
-endif ()
-
-FIND_LIBRARY(METIS_LIB NAMES metis)
-FIND_PACKAGE(Metis ${gunrock_REQUIRED_METIS_VERSION} REQUIRED)
-SET(METIS_FOUND FALSE)
-if (METIS_LIB)
-  set(METIS_FOUND TRUE)
-  add_definitions( -DMETIS_FOUND=true )
-  message(STATUS "Found Metis")
-else (METIS_LIB)
-  add_definitions( -DMETIS_FOUND=false )
-  message(WARNING "Metis was requested but support was not found,
-  run `sudo apt-get install metis` or `/dep/install_metis.sh` for installation.")
-endif (METIS_LIB)
-
-if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-  # also see how to do overrides:
-  # https://gist.github.com/robertmaynard/11297565
-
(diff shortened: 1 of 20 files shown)


jowens:
Hmmmm ... the `-Wno-deprecated-declarations` ... is there any way that you can *only* apply that to the moderngpu build? I think in general it's a useful flag to have, because we can usually fix the issues it finds (but not in moderngpu).

neoblizz:
@jowens I tried, but it's in a compiler flag so it gets applied to everything I believe. The only other alternative is to modify modergpu code.

According to what I read up on and discussed with @yzhwang, we've already switched to c++11 and not gotten any deprecated variable warnings, which tells us that we're already using the most up-to-date stuff in our code (except moderngpu). It is unlikely that we'll add something outdated that will give us the warning again, but the possibility still exists.

> A program element annotated @ Deprecated is one that programmers are discouraged from using, typically because it is dangerous, or because a better alternative exists. Compilers warn when a deprecated program element is used or overridden in non-deprecated code.

jowens:
@neoblizz http://stackoverflow.com/questions/13638408/cmake-override-compile-flags-for-single-files ?

neoblizz:
I'll give that a try!

neoblizz:
I am not getting the desired effect from using override for a single file;
```
set_source_files_properties(${MODERN_GPU_FILES_W} PROPERTIES CMAKE_C_FLAGS -Wno-deprecated-declarations)
set_source_files_properties(${MODERN_GPU_FILES_W} PROPERTIES CMAKE_CXX_FLAGS -Wno-deprecated-declarations)
```

- `${MODERN_GPU_FILES_W}` are the files name that gave the warning message.

A reason that may be making this fail is that once you include the file in one of the test primitive (it is used in a lot of places), the warning flag is not carried over to that new project directory/file. @yzhwang

neoblizz:
Per different suggestions I got, last commit reduces the amount of code per every `tests/<primitive-name>/CMakeLists.txt` files to just three lines of code, example:
```
project(bfs)
message("-- Project Added: ${PROJECT_NAME}")
include(${CMAKE_SOURCE_DIR}/cmake/SetSubProject.cmake)
```
- **Note:** In `project(name)` name has to be the name of executable you're adding, like `test_<name>.cu`.

Neat and efficient!

yzhwang:
Good job @neoblizz ! New cmake framework looks really elegant! My understanding is, now to build a project under tests/primitive_name, you only need to type:
cmake -option={bfs,sssp,dobfs,...}?

On warnings, I agree with @neoblizz . Because we are using moderngpu in our advance operator, and it is used in almost all primitives, so simply adding warning flag for mgpu files is not enough. I'm fine with the current way of handling warnings, since it's only suppress the deprecated declarations. I think it won't give us trouble for now. Another way is to post stackoverflow question as the previous one, I guess people may come up with better solutions :)

neoblizz:
Thanks @yzhwang, I haven't added the option thing in this pull request. This keeps the functionality the same just cleans up a lot of code and separates things with include files. Also have introduced `project(primitive)` in this pull request which I can later use to implement option stuff.

neoblizz:
Addressed deprecated declarations warning by adding a flag that can be turned on if we want to see all the warnings: `cmake -DWARNING_DEP_FLAG=ON ..` where `..` is the location of gunrock.

We can now also build **individual applications** instead of building the entire project with the following command: `cmake -DGUNROCK_BUILD_APPLICATIONS=OFF -DGUNROCK_APP_BFS=ON -DGUNROCK_APP_BC=ON ..`

- `-DGUNROCK_BUILD_APPLICATIONS=OFF` turns off "Build all applications".
- `-DGUNROCK_APP_BFS=ON` allows us to only build bfs.
- `-DGUNROCK_APP_BC=ON` allows us to only build bc.

When executed, it shows what projects were added to the build.
```
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found CUDA: /usr/local/cuda-7.5 (found suitable version "7.5", minimum required is "7.5") 
-- Boost version: 1.58.0
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   system
--   filesystem
--   timer
--   chrono
-- Try OpenMP C flag = [-fopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Success
-- Try OpenMP CXX flag = [-fopenmp]
-- Performing Test OpenMP_FLAG_DETECTED
-- Performing Test OpenMP_FLAG_DETECTED - Success
-- Found OpenMP: -fopenmp  
-- Found OpenMP
-- Found Metis
-- Project Added: bc
-- Project Added: bfs
-- Configuring done
-- Generating done
-- Build files have been written to: /home/muhammad/gunrock/build-cmake
```

@yzhwang ready to merge.

yzhwang:
Thanks @neoblizz ! This looks awesome! Just make sure to record all the changes somewhere so that when we finally merge this into master in our next release, the changelog is accurate. I'm thinking of using github's wiki page to store this. https://github.com/gunrock/gunrock/wiki

jowens:
Why store it on the wiki? Just create a changelog-post-1.3.md and put it there. Then you can copy it into release-notes.md when we release.

yzhwang:
I was thinking of using wiki because it works for all branches (we have changes now on dev, dev_debug, dev-intersection-op, dev_sm). I agree that creating changelog under each branch and then merge back to dev is a more convenient way.

jowens:
Moving to moderngpu 2.0 would actually be the best long-term solution, esp. since we can probably convince Sean to fix any compiler warnings we find there.

neoblizz:
I believe they're using c++11 so this might have already been fixed. But yes, that will be the ideal solution. Looking forward to seeing moderngpu 2.0's features being used.

PR #230, opened 2017-01-07
Fix broken link

arokem:


jowens:
I'm puzzled here because this link works fine for me.

arokem:
Clicking on the link in the README on your Github repo sends me here: 

screen shot 2017-01-06 at 7 55 53 pm

jowens:
Yeah, never anticipated people would go from the github.com page to the links within it (rather than from the github.io page). But of course they would. Anyway, there were lots and lots of links on that page that were broken, not just the one you pointed out. Fixed here, thanks for pointing it out. https://github.com/gunrock/gunrock/commit/8ca8a3fcc4585d991a67f35e7b4dbb12d584990e

PR #234, opened 2017-02-01
Added support for float edge values.

GMolini:
Adds support for float edge values in the SSSP algorithm.

### DIFF
--- a/gunrock/app/sssp/sssp_app.cu
+++ b/gunrock/app/sssp/sssp_app.cu
@@ -324,10 +324,52 @@ float dispatchSSSP(
             }
             case VALUE_FLOAT:
             {
-                // template type = <int, float, int>
-                // not support yet
-                printf("Not Yet Support This DataType Combination.\n");
-                break;
+              Csr<int, int, float> csr(false);
+              csr.nodes = graphi->num_nodes;
+              csr.edges = graphi->num_edges;
+              csr.row_offsets    = (int*)graphi->row_offsets;
+              csr.column_indices = (int*)graphi->col_indices;
+              csr.edge_values    = (float*)graphi->edge_values;
+              parameter->graph = &csr;
+
+              // determine source vertex to start
+              switch (config -> source_mode)
+              {
+              case randomize:
+              {
+                  parameter->src[0] = graphio::RandomNode(csr.nodes);
+                  break;
+              }
+              case largest_degree:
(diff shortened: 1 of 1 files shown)


yzhwang:
Hi @gumlym Thanks for your pull request! I have some suggestions for you:
1) Currently we hard code the value data type in our shared_lib_app for SSSP here:
https://github.com/gunrock/gunrock/blob/master/gunrock/app/sssp/sssp_app.cu#L421
could you make sure that different data type could propagate there?
2) Could you add a command line option for the test in shared_lib_tests for SSSP, so that we could test whether this change will break our implementation?
https://github.com/gunrock/gunrock/blob/master/shared_lib_tests/shared_lib_sssp.c

Thank you so much for making Gunrock more flexible! Let me know if you have any concern. When you finished you can submit another pull request and we will merge that one instead.

gumlym:
Hi,

Theres a problem with the first thing. The sssp interface has also hardcoded the values of distances and edge_values (unsigned int*). If I just add another interface of sssp with those variables as float*, i get an error because gunrock.h then gets linked with C, and C doesnt allow overloading. Im not sure how youd like to proceed. One option is to set those variables as void * (like you do in the GRGraph struct), and just add another parameter ValueType to the function call  where you can set the desired type. That would require changes in the python way of calling the function. Another would be to do a second interface sssp_float with the harcoded values. I can do it whichever way you prefer.

yzhwang:
Thanks @gumlym I didn't notice the sssp interface when I replied to you. It's been a while since I wrote the stuff. I'd prefer the first one, as it gives more flexibility and could serve as a template for the later modifications of other primitives. I'll get @sgpyc into the loop as he is now the primary developer of this project. He is actually doing some refactoring work now. If you have any other suggestions, let us know. Again, thank you so much for helping make Gunrock better!
@sgpyc , could you help @gumlym on this pull request and if your schedule is tight, could you assign a student to do this? I think this is a good opportunity to learn the pipeline in the framework. Thank you!

yzhwang:
A friendly ping of this, we need to find someone to take care of this. Thanks!

jowens:
@neoblizz can you take a look please?

neoblizz:
@jowens on it!

neoblizz:

neoblizz:
Supported in current refactoring.

PR #267, opened 2017-10-07
Gunrock's Forking Workflow.

neoblizz:
Explanation of how we should structure gunrock's development using different forks and pull requests (this pull request being an example).

Sources: https://www.atlassian.com/git/tutorials/comparing-workflows#forking-workflow

### DIFF
--- a/doc/forking_workflow_gunrock.md
+++ b/doc/forking_workflow_gunrock.md
@@ -3,28 +3,30 @@ Git Forking Workflow
 
 Transitioning over from [Git Branching Workflow](http://nvie.com/posts/a-successful-git-branching-model/) suggested by Vincent Driessen at nvie to [Git Forking Workflow](https://www.atlassian.com/git/tutorials/comparing-workflows#forking-workflow) for Gunrock.
 
-##How Forking Workflow Works?
-
+How Forking Workflow Works?
+=============
 ![Forking Workflow](https://wac-cdn.atlassian.com/dam/jcr:5c0941ff-a8b5-435b-a092-2167705f1e97/01.svg?cdnVersion=hp)
 > As in the other Git workflows, the Forking Workflow begins with an **official public repository** stored on a server. But when a new developer wants to start working on the project, they do not directly clone the official repository.
 <br><br>Instead, they **fork the official repository** to create a copy of it on the server. This new copy serves as their personal public repository—no other developers are allowed to push to it, but they can **pull changes** from it (we’ll see why this is important in a moment). After they have created their server-side copy, the developer performs a git clone to get a copy of it onto their local machine. This serves as their private development environment, just like in the other workflows.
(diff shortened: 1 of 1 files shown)


neoblizz:
Doc: https://github.com/neoblizz/gunrock/blob/master/doc/forking_workflow_gunrock.md

ctcyang:
@sgpyc mentioned that one of the reasons we haven't been doing this is because it is time-consuming to manually run comprehensive test to check that the PR didn't break anything. What are some tools that people use to automate testing whenever someone submits a PR?

neoblizz:
It is very common to use a bot to do that, and it can report results to slack. But yeah, having someone who is testing the code is also part of this (as an added check). @ctcyang idk if this is the answer, but I have seen projects use https://travis-ci.org/ and https://codecov.io/.

yzhwang:
We can set up Jenkins too. https://github.com/jenkinsci/jenkins

jowens:
So the big difference between the current workflow and this workflow is that development branches would no longer be branches but instead be forks. Accurate? What's the advantage here? I do like the ability to look at a pull request and comment on it.

I think CI would be awesome. Not clear to me if we can use a cloud service (in which case I don't know which has GPU support, but I know NVIDIA is working on it) or if we need to set up something locally (in which case I'd be happy to ask COE if they'd set one up). Thoughts?

neoblizz:
Yes, that is accurate. Having things in forks and then having the pull requests for the changes encourages code review, or at the very least some automated testing and instant feedback on those using things like Jenkins or Travis CI. In the end, this also simplifies how we push out a release, instead of merging tons of branches we only have to test one dev branch, because that's where all the changes will go to now (less work for whoever is doing the cleanup before the release).

I read somewhere that Travis CI was interested in supporting GPUs for CUDA. I'll have to check up on that again, but if their Cloud service doesn't work for us, we can host one even on mario or COE as you suggested.

neoblizz:
Should I start setting up Jenkins? @jowens Should we go ahead with this approach or are there any other obstructions/questions?

jowens:
I sent a query to COE staff, no reply yet.

First, do we want Jenkins or Travis CI? 

I think it's probably most sensible to start with picking a CI system (Jenkins or Travis CI) then next writing a test before we set up the CI system. I don't want to see you make the effort to set up a CI system then we find something outside the group we can use instead. But I know we need to write tests.

neoblizz:
I am working on setting up gtest with a sample test, but Jenkins or Travis CI can take one of our current gunrock apps tests written already.

jowens:
Also, transition plan? We have several long-lived branches. Be specific about how we transition. Presumably branch owner checks in, has no more branches, then does further development with new plan?

neoblizz:
I am creating an issue per branch and assigning it to the branch owner. I would request for them to either create a new fork or update their personal fork so that they have all the branches existing on their fork before the branch is removed. This way, all they have to do is create a pull request from their feature branch to the dev branch once it is time to merge.

PR #273, opened 2017-10-25
Incorporating Unit Testing using Googletest (DO NOT MERGE).

neoblizz:
This pull request introduces Googletest to Gunrock. Allowing us to write unit tests, I am currently in the process of writing a unit test as a sample for the future ones. I have added Googletest as one of the gunrock's external projects (submodule) to be used by the `gunrock/test` files. Also moved all the `shared_lib` under `tests` as well. 

- [x] Add googletest as an external project.
- [x] Move all present testing routines under the the testing layer.
- [x] Introduced googletest to CMakeLists.txt and also created a new cmake file for tests.
- [x] Create a sample unit test (I am doing so by leveraging the shared_lib_bfs, and converting it into a gtest instead).
- [x] Write a reference code for the ASSERT statement.


neoblizz:
@jowens ok, good, because I was worried about it not being able to support 2.8 or lower. Adding an external project is far easier in later cmake versions. But the tests now run beautifully:

```
muhammad@blizzard:~/temp/gunrock/build/bin (master)*$ ./unit_test 
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from sharedlibrary
[ RUN      ] sharedlibrary.breadthfirstsearch
Round 1 of bfs.
0	 1	 0	 queue3 	 oversize :	 9 ->	 10
Round 2 of bfs.
0	 1	 0	 queue3 	 oversize :	 10 ->	 11
Round 3 of bfs.
[       OK ] sharedlibrary.breadthfirstsearch (6565 ms)
[----------] 1 test from sharedlibrary (6565 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (6565 ms total)
[  PASSED  ] 1 test.
```

Googletest is neat!

jowens:
I think we should be aggressive in not supporting older stuff if newer stuff is working for us (e.g., dropping support for older CUDAs in favor of C++11).

Is there a tool that shows code coverage for googletests?

I think we would benefit from a little time from you (or someone else) in pointing to places where we would benefit from googletests. You're actually somewhat familiar with how they should be used. Issues might be good, even if they're unassigned issues? I don't want redundant tests but we should think about which tests would be most useful in the near term.

tests/unit_tests/test_lib_bfs.h
```diff
@@ -34,8 +35,17 @@ TEST(sharedlibrary, breadthfirstsearch)
...
+    // int node; for (node = 0; node < graphi->num_nodes; ++node)
+    //    printf("Node_ID [%d] : Label [%d]\n", node, labels[node]);
+
+    int result[7] = {2147483647, 2147483647, 0, 1, 1, 1, 2};
+
+    // ASSERT_THAT(labels, ElementsAre(2147483647, 2147483647, 0, 1, 1, 1, 2)) << "BFS using Gunrock as a shared library was unsuccessful";
+    // ASSERT_EQ(labels.size(), result.size()) << "Vectors x and y are of unequal length";
+
+    for (int i = 0; i < graphi->num_nodes; ++i) {
+      EXPECT_EQ(labels[i], result[i]) << "Vectors x and y differ at index " << i;
```
jowens: @neoblizz this is the actual test, which is "compare the generated labels to the expected result"?

neoblizz:
That's what I had in mind next, I would like to get the bots for code coverage and cl working for the pull requests. I also plan on creating issues per unit tests, so if any of the new students would like to add to gunrock by picking them, it will be of great help!

tests/unit_tests/test_lib_bfs.h
neoblizz: Yes. `EXPECT` means the test will continue even if it fails, if I used an `ASSERT` there it will be a "fatal" test. Those are the only two tests that are used, with different operators (for example NE, LT, GT, etc.)

neoblizz:

neoblizz:
Did some major reordering of things, and codecoverage and gtests now works perfectly. The only change that counts from the user's perspective is now `ctest` command will work under `/path/to/build/directory/examples` instead of `/path/to/build/directory`. @sgpyc could you please take a look at this and maybe we can consider a merge?

jowens:
Can we have a top-level ctest that just calls all the ctests in the examples directory (or calls the one ctest there)?

neoblizz:
I will look into it. I think I can do that with `make check` but I am not sure if I can recursively run `ctest`.

Issue #261, opened 2017-09-14
Load Graph Error: Invalid Graph (MTX Required)


neoblizz:
Discussed in Issue #260, if an invalid graph exit with a proper error message instead of segfaults.

neoblizz:
Handled in the PR above.

PR #274, opened 2017-10-26
Added check for invalid graph (exits if number of nodes is negative).

shivampdesai:
Use of gdb indicated that invalid graphs generated a negative number of nodes, which caused a segmentation fault. This check will display an error message if this is the case and terminate the program. 

This can be tested with the simple graph:
2 2 2
1 1
2 2

which used to seg fault but now will cause an exit out of the program.

### DIFF
--- a/gunrock/graphio/market.cuh
+++ b/gunrock/graphio/market.cuh
@@ -363,7 +363,8 @@ int ReadMarketStream(
 
 		if (ll_row < 1 || ll_col < 1)
                 {
-                    fprintf(stderr, "\nError Invalid graph: Bad MTX format, indices cannot start with 0.\n");
+                    fprintf(stderr, "\nError Invalid graph: Bad MTX format, 
+			    indices cannot start with 0.\n");
                     if (coo) free(coo);
                     exit(1);
                 }
--- a/gunrock/graphio/market.cuh
+++ b/gunrock/graphio/market.cuh
@@ -71,14 +71,6 @@ int ReadLabelStream(
                 return -1;
             }
 
-            if (ll_nodes_x != ll_nodes_y)
-            {
-                fprintf(stderr,
-                        "Error parsing node labels: not square (%lld, %lld)\n",
-                        ll_nodes_x, ll_nodes_y);
-                return -1;
-            }
-
             nodes = ll_nodes_x;
 
             if (!quiet)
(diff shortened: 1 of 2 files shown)


neoblizz:

neoblizz:
@shivampdesai this addresses one of the issues, I think the big portion of this project was getting a system that recognizes if true MTX is passed to gunrock, and if not it should print out an error. Would you consider tackling this? https://github.com/gunrock/gunrock/issues/261

shivampdesai:
Yes. What are the characteristics that would define a file in MTX format (things that I can check for)?

neoblizz:
@shivampdesai it is all right here: http://math.nist.gov/MatrixMarket/formats.html

shivampdesai:
free(coo) is done in both cases and I updated it so that it is less than 80 characters.

gunrock/graphio/market.cuh
```diff
@@ -327,7 +327,8 @@ int ReadMarketStream(
 
 		if (ll_row <  1 || ll_col <  1)
 		{
-		    fprintf(stderr, "\nError Invalid graph: Bad MTX format, indices cannot start with 0.\n");
+		    fprintf(stderr, "\nError Invalid graph: Bad MTX format, indices
```
jowens: Pretty sure you don't want a newline in the middle of a string. I think what you want is

```
fprintf(stderr, "start of text "
        "next line of text");
```

gunrock/graphio/utils.cuh
```diff
@@ -65,8 +65,7 @@ SizeT RandomNode (SizeT num_nodes)
     return node_id % num_nodes;
 }
 
-template <typename VertexId, typename SizeT, typename Value>
-void RemoveStandaloneNodes(
+template <typename VertexId, typename SizeT, typename Value>void RemoveStandaloneNodes(
```
jowens: not sure why this change is here -- put a newline between `Value>` and `void`

gunrock/graphio/market.cuh
```diff
@@ -332,6 +324,15 @@ int ReadMarketStream(
             {
                 num_input = sscanf(line, "%lld %lld %lf",
                     &ll_row, &ll_col, &lf_value);
+
+		if (ll_row <  1 || ll_col <  1)
+		{
+		    fprintf(stderr, "\nError Invalid graph: Bad MTX format, indices
+			    cannot start with 0.\n");
+		    if (coo) free(coo);
```
jowens: I still see an `if (coo)` here ...

gunrock/graphio/market.cuh
```diff
@@ -346,8 +347,7 @@ int ReadMarketStream(
 
                 else if (array || num_input < 2)
                 {
-                    fprintf(stderr,
-                            "Error parsing MARKET graph: badly formed edge\n");
+                    fprintf(stderr,                            "Error parsing MARKET graph: badly formed edge\n");
                     if (coo) free(coo);
```
jowens: I still see an `if (coo)` here ...

PR #280, opened 2017-11-03
(PR#273 Continued) Googletest/Unittests, Codecove, Jenkins.

neoblizz:
Continuing from Pull Request #273, this is same cherry-picked commits now introduced to dev branch.


neoblizz:
@sgpyc this is the same pull request to `dev` branch, I have tested it and it works. Could you do a code review and see if it is ready to be merged?

tests/unit_tests/test_lib_bfs.h
```diff
@@ -35,14 +35,8 @@ TEST(sharedlibrary, breadthfirstsearch)
 
     int *labels = (int*)malloc(sizeof(int) * graphi->num_nodes);
     labels = (int*)grapho->node_value1;
-    // int node; for (node = 0; node < graphi->num_nodes; ++node)
-    //    printf("Node_ID [%d] : Label [%d]\n", node, labels[node]);
-
     int result[7] = {2147483647, 2147483647, 0, 1, 1, 1, 2};
```
neoblizz: @jowens Should I be writing a reference BFS or having hard coded results work? I was looking at how `tinydnn` implemented gtest and for most of their stuff they had a file with manually written results. That's why I went for this. Your thoughts?

cmake/DownloadProject/README.md
```diff
@@ -0,0 +1,24 @@
...
+in the article available at:
+
+https://crascit.com/2015/07/25/cmake-gtest/
+
+An example as described in that article is provided here to demonstrate
+how to use the DownloadProject module. It uses [googletest][1] as the
+example, downloading and building trivial gtest and gmock test cases
+to show the technique.
+
+[1]: https://github.com/google/googletest
```
neoblizz: @jowens note that I didn't write `cmake/DownloadProject`, is this ok for me to do? We have used cmake modules from others before.

jowens: It's got an MIT license: https://github.com/Crascit/DownloadProject/blob/master/LICENSE

Just make sure you follow the license terms.

sgpyc:
I saw some mtx files from `tests/astart` and `tests/lp` have been removed? shall we keep them in `datasets/small`?

tests/test_lib_bfs.h
```diff
@@ -0,0 +1,49 @@
+/**
```
sgpyc: should this file be under `tests/bfs`?

neoblizz:
Yes, I moved `bips98_606.mtx` to `dataset/small`, but didn't intend to remove those `.mtx` files. I think the merge got rid of them by accident. We can move all those datasets under small.

tests/test_lib_bfs.h
neoblizz: If we would like to further have subdirectories for gtests I can move that under tests/bfs, right now since it is the only unit test we have I just kept it under tests.

neoblizz:
Added the `mtx` files back in to `dataset/small` and moved the test under `tests/bfs`.

PR #292, opened 2017-11-16
Check for empty first line in input file stream.

shivampdesai:
According to MTX format guidelines, the first line of the file cannot be empty. Checks for empty first line, displays error message and exits out if it is empty. Can be tested by adding an empty line above any simple graph, for example:

4 4 8
2 1
3 1
4 1
3 2
4 2
1 3
1 4
3 4

will generate an error and exit out of the program.


shivampdesai:
Yes, an empty line causes a break from the loop and passes coo as a null pointer, which causes an error outputted as "No graph found".

neoblizz:
I think this is great, even if some MTX file has an empty line, it prints out an error indicating what's wrong with the graph. It is not too much to ask someone to get rid of the first line of their MTX file per the official guidelines.

PR #310, opened 2018-02-13
Check for LB_CULL traversal mode

shivampdesai:


neoblizz:
@shivampdesai could you please confirm that when you set traversal mode = LB_CULL it prints an error and exits?

shivampdesai:
Yes, when traversal_mode is set to "LB_CULL" (which is hard-coded in to test), it will print out the error: "Traversal Mode LB_CULL not available for PageRank" and exit out of it.

neoblizz:
Perfect, thanks. Lets run the CI tests.

PR #315, opened 2018-02-18
[FIX] CMake warnings under OS X: MACOSX_RPATH is not specified

### LINKED ISSUE #316: CMake warnings under OS X
CMake warnings under OS X: MACOSX_RPATH is not specified for the following targets.

gangliao:
fix #316


neoblizz:
Your latest PR took care of this commit as well.

PR #325, opened 2018-03-10
Workaround for the Volta + CUDA 9.1 bug

sgpyc:
Adding two _all(1) to sync threads within a warp keeps WarpScan running


neoblizz:
@sgpyc do you plan on reverting this after the bug's been resolved?

sgpyc:
Should keep the active masks for warp scans and the input_queue_length handling parts, only the following two lines should be removed after the bug is been resolved (maybe plus some clean-ups):
https://github.com/gunrock/gunrock/pull/325/files#diff-5ef2d68740d6fa2f862cbe119651e545R86
https://github.com/gunrock/gunrock/pull/325/files#diff-5ef2d68740d6fa2f862cbe119651e545R169

neoblizz:
Ok, understood. This will be good to document here in case we need to revisit this. Merging.

PR #326, opened 2018-03-11
Merge master commits back to dev

sgpyc:


sgpyc:
Do not merge this, use #327 instead

PR #339, opened 2018-05-22
[Louvain] CPU/Omp implementation, GPU halfway 

sgpyc:


neoblizz:
For just CPU implementation:
> commit 10e2412 is the last compilable commit

PR #346, opened 2018-06-29
Cleaning up the scripts folder, expand on [standard] directory.

neoblizz:


neoblizz:
Doesn't break anything, just cleaned up some scripts from topc and created a new folder for it.

PR #351, opened 2018-07-01
Updating CUDA Architecture properties.

neoblizz:
Check out the last commit in this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,8 @@ set(GENCODE_SM37
   -gencode=arch=compute_37,code=sm_37 -gencode=arch=compute_37,code=compute_37)
 set(GENCODE_SM50
   -gencode=arch=compute_50,code=sm_50 -gencode=arch=compute_50,code=compute_50)
+set(GENCODE_SM52
+  -gencode=arch=compute_52,code=sm_52 -gencode=arch=compute_52,code=compute_52)
 set(GENCODE_SM60
   -gencode=arch=compute_60,code=sm_60 -gencode=arch=compute_60,code=compute_60)
 set(GENCODE_SM61
@@ -214,7 +216,7 @@ option(GUNROCK_GENCODE_SM30
   OFF)
 
 option(GUNROCK_GENCODE_SM35
-  "ON to generate code for Compute Capability 3.5 devices (e.g. Tesla K20)"
+  "ON to generate code for Compute Capability 3.5 devices (e.g. Tesla K40)"
   ON)
 
 option(GUNROCK_GENCODE_SM37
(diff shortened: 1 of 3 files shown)


jowens:
Thanks for sharing this. Now I guess it's a question of _using_ these new properties you've exposed. :)

neoblizz:
In that regards, Gunrock automatically calculates the best launch parameters within kernel policies. Hope is, my next runs reflect that. Just to give an example on this:

https://github.com/gunrock/gunrock/blob/f011bb033669091a8c3a6c6c3b0910e20287c615/gunrock/oprtr/edge_map_partitioned/kernel_policy.cuh#L102

jowens:
Right, trying to figure out how that's actually used. Eventually there's a computation where there's a ... computed minimum number of threads per block, based on smem usage, max threads per block, etc.? and we launch based on that? What's the goal here? Compute the right number of threads per block to launch? And we've been launching too few?

neoblizz:
Yes, we do that computation and launch enough to fill the GPU for a given job. We've been launching enough to easily fill a Kepler, but hit a wall when it came to Pascal and Volta because the hardware got better.

Another kind of tuning on top of that is how many blocks we should launch for a given job, right now that number is 2. What Yuechao mentioned yesterday was it might be worth revisiting these numbers to see if they are still the optimal for us in Volta.

jowens:
Correct me if I'm wrong, since I want to understand this.

There are two things we're trying to set here. One is threads per block. One is simultaneously running blocks per SM. They're obviously related. I'm presuming the underlying code works with a varying number of threads per block.

All other things being equal, I'm presuming having more threads per block is better (we get a performance benefit).

Which is the "independent' variable here? Which do we want to tune? If we set one, it kind of sets the other, yes?

neoblizz:
Threads per block is fixed, we launch as many as we can per block.
And we launch exactly 2 blocks per SM (so that is set to in a way).

It is worth playing with these two together, to see what works best. Because they are obviously related.

Right now, we limited ourselves to only use 46KB of shared memory, we were wasting half of it in Volta and around 16KB in Pascal. Messing around with number of blocks, and threads per blocks allows us to redo our past findings of 2 blocks and 1024 threads per block.

There's also odd arch with 4096 threads (the 6.2 arch I think) that we don't consider. That is P104 if I remember correctly.

neoblizz:
Note on shared memory, is that max that you can have per block is still 46KB, but in an SM is 96KB for V100 and 64KB in P100. Now, V100 is obvious, you launch 2 blocks with 46KB each, but P100 could be a different split.

Same thing goes for the registers in use.

jowens:
Are these parameters in the JSON? I'm presuming the one we actually need to record is threads per block.

neoblizz:
Hm, no. We should definitely record these if possible. Right now, the utilization is studied through the profiler. Only thing we have in JSON is the gpuinfo:

```
    "gpuinfo" : {
        "clock_rate" : 1380000,
        "compute_version" : 70,
        "driver_api" : 9010,
        "driver_version" : 9010,
        "major" : 7,
        "minor" : 0,
        "multi_processor_count" : 80,
        "name" : "Tesla V100-PCIE-32GB",
        "runtime_version" : 9010,
        "total_global_mem" : 34089730048
    }
```

jowens:
It's time-consuming because of the rebuild, but doing a full run on Volta for each of a varying set of launch parameters is a good way to start exploring what the right parameters are.

neoblizz:
Yes, I intend to do that! It is worth exploring.

PR #354, opened 2018-07-06
Integrating docs in main repo and benchmarking script improvements...

neoblizz:


scripts/standard/run-tests.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
+TIMESTAMP=`date '+%Y-%m-%d_%H:%M:%S'`
+
 EXEDIR=${1:-"../../build/bin"}
 DATADIR=${2:-"/data/gunrock_dataset/large"}
 DEVICE=${3:-"0"}
-TAG=${4:-"V100"}
+TAG=${4:-"$TIMESTAMP"}
```
jowens: This is a good idea. I presume the same tag goes to every experiment in the same run. I like it.

neoblizz:
@jowens needs a review for last commit, the idea is to reduce duplicates and move everything to `docs` except the doxygen files. And I have renamed the `doc` to `doxygen`.

jowens:
Guess it's reasonable, everything in docs is good, it would be good to have some guidance on how we should set things up locally now (previously I had parallel "gunrock" (from gunrock/gunrock) and "gunrock-docs" (from gunrock/docs) directories, that's no longer the case? This should probably be documented somewhere ("best practices to set up your directories for doing Gunrock development").

PR #355, opened 2018-07-07
Updating README to match index.html.md.

neoblizz:
README.md is the only file that will be required to be updated, this is what we were talking about @jowens. Could you please look at this and merge in (or let me know what needs to be changed)? After this, I will push in the gunrock/docs update so it takes in this README for index.html.


neoblizz:
Markdown Tables+Style needs fixing sometime later on. Merging in for now.

PR #364, opened 2018-08-01
[docs] adding quickstart code

bkj:
Add code snippet for build to top of README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,17 @@ Gunrock is featured on NVIDIA's list of [GPU Accelerated Libraries](https://deve
 </tr>
 </tbody></table>
 
+## Quickstart
+```
+git clone --recursive https://github.com/gunrock/gunrock/
+cd gunrock
+mkdir build
+cd build
+cmake ..
+make -j$(nproc)
+make test
+```
+
 ## Gunrock Source Code
 
 <table style="font-size: 12px;"><thead>


README.md
```diff
@@ -31,6 +31,17 @@ Gunrock is featured on NVIDIA's list of [GPU Accelerated Libraries](https://deve
 </tr>
 </tbody></table>
 
+## Quickstart
+```
+git clone --recursive https://github.com/gunrock/gunrock/
+cd gunrock
```
jowens: Would like the opinion of @sgpyc and @neoblizz : I usually make a parallel directory to "gunrock" called "gunrock-build" (instead of putting build within gunrock). Also "gunrock-docs" and "gunrock-io". @bkj is the norm to do a build within the source directory?

bkj: Different people do it differently I think.  I was following the install instructions from [the docker install](https://github.com/gunrock/gunrock/blob/master/docker/ubuntu16.04-cuda9.0/dockerfile)

neoblizz: I create a build directory inside of gunrock, and this is how Jenkins tests it as well. @sgpyc creates a parallel directory to gunrock. :)

One thing though is that do you we wanna put this under docs instead of individual folders within our main repo? For example, one page on "gunrock's docker guide"?

bkj: Either way is fine, but it's probably good to be consistent.  I think `build` inside the repo is more standard than a parallel folder?  I don't have a strong opinion though.

Not sure I understand your second question?

jowens: "more standard" is my preference.

Issue #370, opened 2018-08-14
[branch:dev-refactor] Call to `BaseDataSlice::Init`


bkj:
Calling `BaseDataSlice::Init` like
```
GUARD_CU(BaseDataSlice::Init(sub_graph, gpu_idx, target, flag));
```
as in the template [here](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/app/Template/Template_problem.cuh#L134) appears to get parsed incorrectly [here](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/app/problem_base.cuh#L98): `gpu_idx` gets assigned to `num_gpus`, etc.  

I think the line in the template is supposed to look like [this](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/app/pr/pr_problem.cuh#L229), where `num_gpus` is passed into `DataSlice::Init` and then into `DataSliceBase::Init`.

Removing the default arguments in `DataSliceBase::Init` would help avoid this problem by throwing a compiler error if the number of arguments is incorrect.

~ Ben

jowens:
@sgpyc can you comment on this please?

neoblizz:
You are right, I can issue a fix for this.

> Removing the default arguments in DataSliceBase::Init would help avoid this problem by throwing a compiler error if the number of arguments is incorrect.

I have had some issues with this as well, having default parameters could cause confusion specially when a lot of the stuff is templated. I think on the other hand, it is also a common practice and handy sometimes to expect from user to not type in literally every parameter if it is a simple NULL, 0 or a default value. I am iffy about this.

sgpyc:
Parameter matching is terrible, and I spent quite sometime doing that, but couldn't catch all the mismatches.
I really hope C/C++ can have named parameters, but...
A workaround is using a structure to hold some parameters (like what oprtr_parameters does). But it will be too much if using parameter structures for a large number of functions.

If anything works better, I would love to know and try

bkj:
Should I open a quick PR for this?

neoblizz:
@bkj yes please!

neoblizz:
Solved with PR #381, thanks @bkj!

PR #380, opened 2018-08-30
[dev-refactor] Improving simple template for refactored API.

bkj:
This is another version of the `tests/Template` and `gunrock/app/Template`.  Not meant to replace that code, but as another example for someone new to Gunrock.

- removed some code from the `tests/Template` that didn't seem totally general. 
- use `<todo>...</todo>`  tags to indicate TODOs -- the `//TODO` format caused a few minor points of confusion when I was implementing `BC`
- implements a working "hello world" app that computes the degree of all nodes -- having a working program that's as simple as possible can be helpful to novices.  (`tests/sssp` is fairly starightforward, but this is even simpler)

I haven't fully fleshed out the template for multi-GPU apps yet.


neoblizz:
I think one thing that we need to do is separate the word "Template" (often used for C++ templates) from this "sample" application. We could call it "hello-world", "hello-gunrock" or just "sample", anything will work better than overloading the word "template".

gunrock/app/SimpleTemplate/SimpleTemplate_app.cu
```diff
@@ -0,0 +1,341 @@
...
+// <todo> change includes
+#include <gunrock/app/SimpleTemplate/SimpleTemplate_enactor.cuh>
+#include <gunrock/app/SimpleTemplate/SimpleTemplate_test.cuh>
+// </todo>
+
+namespace gunrock {
+namespace app {
+// <todo> change namespace
+namespace SimpleTemplate {
+// </todo>
```
neoblizz: I really like the use `<todo>`, one note though is that using `TODO` in a comment in most text editors is actually highlighted in different colors, even in editors like vim it shows up really nice and bright. But maybe having it open and close with `<todo></todo>` tag helps.

gunrock/app/SimpleTemplate/SimpleTemplate_app.cu
```diff
@@ -0,0 +1,341 @@
...
+/**
+ * @brief Run SimpleTemplate tests
+ * @tparam     GraphT        Type of the graph
+ * @tparam     ValueT        Type of the distances
+ * @param[in]  parameters    Excution parameters
+ * @param[in]  graph         Input graph
+...
+ * @param[in]  target        where to perform the app 
+ * \return cudaError_t error message(s), if any
+ */
```
neoblizz: @jowens has noted before that we need to be more detailed with commenting the parameters.

gunrock/app/SimpleTemplate/SimpleTemplate_enactor.cuh
```diff
@@ -0,0 +1,417 @@
...
+        
+        // <todo> add problem specific data alias here:
+        auto &degrees = data_slice.degrees;
+        auto &visited = data_slice.visited;
+        // </todo>
+        
+        // --
+        // Define operations
+
+        // advance operation
```
neoblizz: Lets describe advance operator here.

gunrock/app/SimpleTemplate/SimpleTemplate_enactor.cuh
```diff
@@ -0,0 +1,417 @@
...
+            // Increment degree of src
+            atomicAdd(degrees + src, 1);
+            
+            // Add dest to queue if previously unsen
+            return dest_visited == 0;
+            
+            // </todo>
+        };
+
+        // filter operation
```
neoblizz: Lets describe filter operator here.

gunrock/app/SimpleTemplate/SimpleTemplate_enactor.cuh
```diff
@@ -0,0 +1,417 @@
...
+            // <todo> implement filter operation
+            return true;
+            // </todo>
+        };
+        
+        // --
+        // Run
+        
+        // <todo> some of this may need to be edited depending on algorithmic needs
+        // !! How much variation between apps is there in these calls?
```
neoblizz: Not a lot of variance in the calls I think. Only thing that makes sense is optimizations like a fused kernel call instead of two separate ones.

neoblizz:
@bkj we can use this PR to address issue #370.

bkj:
I can change the name from template to "hello", and then we can add a README to the `tests/` directory that points people at the appropriate "templates"

bkj:
Pushed those changes.

TODO:
 - [ ] Add some documentation about advance/filter ops in `hello_enactor.cuh`

jowens:
I had a speck of dust on my screen and read "hello" as "hella" and thought, wow, @bkj is totally from NorCal now.

bkj:
@neoblizz can we merge this?  are there other changes you'd like to see made?

PR #383, opened 2018-08-31
`--random-edge-values` flag

bkj:
Adding `--random-edge-values` flag to change the default behavior when edge weights are not specified in the input file.  Previously, Gunrock generated random edge weights by default, which was surprising behavior.  This PR changes the default behavior s.t. edge weights are set to 1 by default.  The old behavior can be exposed via the `--random-edge-values`  flag.


neoblizz:
This looks good to me, we should probably document this somewhere. Maybe in the parameter description itself say (default edge weights = 1)?

bkj:
Added docs about default behavior

neoblizz:
Thank you!

PR #378, opened 2018-08-28
[dev-refactor] Betweenness Centrality port to new API.

bkj:
Betweenness centrality app using 0.5 API

PR #384, opened 2018-09-04
[dev-refactor] HIVE VN application

bkj:
#### Vertex nomination (VN) kernel

For each node in the graph, compute the minimum distance to a node in the source set.  This is equivalent to SSSP, but w/ multiple nodes in the initial frontier.

Running this app w/ `--src a,b,c` uses nodes `a`, `b` and `c` as the source set.  This is slightly different from SSSP, which runs SSSP from `a`, then from `b`, then from `c`.  If we can figure out a way to pass _sets_ of nodes via `src`, I can make the appropriate changes -- this could possibly look like 
```
--src a+b+c,d+e+f
```
to pass sets `{a,b,c}` and `{d,e,f}`.

To see the difference between `vn` and `sssp`, you can look at this diff: 
    https://github.com/bkj/gunrock/commit/0e4e94969b012f7dd6191856490b4064cffb17d9

The differences are not particularly large, but since `sssp` is a canonical graph operation, I chose to implement this in a separate application.

##### Notes
 - single-GPU implementation only
 - need a C-interface w/o templates?

##### Questions
 - in general, what is the use case for `num_runs > 1` in `*_app.cu` ?

### DIFF
--- a/gunrock/app/vn/vn_app.cu
+++ b/gunrock/app/vn/vn_app.cu
@@ -6,9 +6,9 @@
 // ----------------------------------------------------------------------------
 
 /**
- * @file sssp_app.cu
+ * @file vn_app.cu
  *
- * @brief single-source shortest path (SSSP) application
+ * @brief single-source shortest path (vn) application
  */
 
 #include <gunrock/gunrock.h>
@@ -22,12 +22,12 @@
 #include <gunrock/app/test_base.cuh>
 
 // single-source shortest path includes
-#include <gunrock/app/sssp/sssp_enactor.cuh>
-#include <gunrock/app/sssp/sssp_test.cuh>
+#include <gunrock/app/vn/vn_enactor.cuh>
+#include <gunrock/app/vn/vn_test.cuh>
 
 namespace gunrock {
 namespace app {
-namespace sssp {
+namespace vn {
 
 cudaError_t UseParameters(util::Parameters &parameters)
 {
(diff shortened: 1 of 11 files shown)


jowens:
`num_runs`: is this how many times to run the application from start to finish? In general, the first run of a GPU program is relatively slow since there's a lot of runtime compilation (which is then cached), so the GPU community usually does a bunch of runs and averages the results (possibly ignoring the runtime of the very first run). @sgpyc ?

jowens:
AFAICT there's no standard for specifying a list of sets, so I don't have a preference except that I hope it uses as few must-backslash characters as possible. `{}` would require backslashes, I think, but that's cool, we can go with that.

bkj:
@neoblizz Which issues?  Yes I did copy from SSSP and made changes -- are there errors in SSSP?

gunrock/app/vn/vn_app.cu
```diff
@@ -0,0 +1,313 @@
...
+    
+    parameters.Set("srcs", srcs);
+
+    bool quiet = parameters.Get<bool>("quiet");
+    GraphT graph;
+    // Assign pointers into gunrock graph format
+    graph.CsrT::Allocate(num_nodes, num_edges, gunrock::util::HOST);
+    graph.CsrT::row_offsets   .SetPointer(row_offsets, gunrock::util::HOST);
+    graph.CsrT::column_indices.SetPointer(col_indices, gunrock::util::HOST);
+    graph.CsrT::edge_values   .SetPointer(edge_values, gunrock::util::HOST);
```
neoblizz: 293-295: Need the size parameter in the center, right now 1 is being passed as size; use num_nodes, num_edges, etc as you did in the hella_app.cu.

neoblizz: https://github.com/gunrock/gunrock/pull/378#discussion_r214112605

bkj: Patched the SSSP issue here: https://github.com/gunrock/gunrock/pull/385

bkj: Updated `vn_app.cu` in `bkj/vn`

neoblizz:
Thank you! I think it is ok to keep `FromCSR` if `self` is also `csr`, but lets just merge this in for now. I will do some more testing to see if everything is functional.

PR #386, opened 2018-09-06
[dev-refactor][HIVE] random walk app

bkj:
## Random Walk App

This coarsely follows the implementation on the dev branch:
 - https://github.com/gunrock/gunrock/tree/dev/gunrock/app/rw

This _does not_ follow the advance/filter paradigm, so the `Enactor::Core` function looks substantially different than `sssp`, `vn`, `hello` or `bc`. It just has a call to `ForAll`, which records the the current node in the walk and replaces the elements of the queue with a (uniformly sampled) random neighbor.

`pygunrock` implementation available here:
 - https://github.com/bkj/pygunrock/blob/master/apps/rw.py

To see differences between this implementation and the `hello` app, look at this commit:
 - https://github.com/bkj/gunrock/commit/5730703efbe1f9a52d3218bd0ebb9e023d3e166a
 
### TODO
 - [ ] Implement python + C interfaces
 - [ ] Add multiGPU implementation
 - [ ] Correctness checking
  - Need to ensure that all edges in the random walk are real edges.  How could we test that it's really uniformly distributed?

### Questions
 - Is this a sensible way to implement an app that does not follow the advance/filter model?
 - The implementation on `dev` has `Sorted*` functions -- do we need these?  @neoblizz  thinks they're for load balancing -- is load balancing automatically handled by using `ForAll`?

### DIFF
--- a/gunrock/app/rw/rw_app.cu
+++ b/gunrock/app/rw/rw_app.cu
@@ -43,6 +43,13 @@ cudaError_t UseParameters(util::Parameters &parameters)
          10,   
          "vertex id", 
          __FILE__, __LINE__));
+
+    GUARD_CU(parameters.Use<int>(    
+         "seed",   
+         util::REQUIRED_ARGUMENT | util::OPTIONAL_PARAMETER,  
+         time(NULL),   
+         "seed for random number generator", 
+         __FILE__, __LINE__));
     // </DONE>
 
     return retval;
--- a/gunrock/app/rw/rw_app.cu
+++ b/gunrock/app/rw/rw_app.cu
@@ -1,0 +1,341 @@
+// ----------------------------------------------------------------------------
+// Gunrock -- Fast and Efficient GPU Graph Library
+// ----------------------------------------------------------------------------
+// This source code is distributed under the terms of LICENSE.TXT
+// in the root directory of this source distribution.
+// ----------------------------------------------------------------------------
+
+/**
+ * @file hello_app.cu
+ *
+ * @brief Simple Gunrock Application
+ */
+
+#include <gunrock/gunrock.h>
+#include <gunrock/util/test_utils.cuh>
+#include <gunrock/graphio/graphio.cuh>
+#include <gunrock/app/app_base.cuh>
+#include <gunrock/app/test_base.cuh>
+
+// <TODO> change includes
+#include <gunrock/app/hello/hello_enactor.cuh>
+#include <gunrock/app/hello/hello_test.cuh>
+// </TODO>
+
+namespace gunrock {
+namespace app {
+// <TODO> change namespace
+namespace hello {
+// </TODO>
+
+
+cudaError_t UseParameters(util::Parameters &parameters)
+{
+    cudaError_t retval = cudaSuccess;
+    GUARD_CU(UseParameters_app(parameters));
+    GUARD_CU(UseParameters_problem(parameters));
+    GUARD_CU(UseParameters_enactor(parameters));
+
+    // <TODO> add app specific parameters, eg:
+    // GUARD_CU(parameters.Use<std::string>(
+    //    "src",
+    //    util::REQUIRED_ARGUMENT | util::MULTI_VALUE | util::OPTIONAL_PARAMETER,
+    //    "0",
+    //    "<Vertex-ID|random|largestdegree> The source vertices\n"
+    //    "\tIf random, randomly select non-zero degree vertices;\n"
+    //    "\tIf largestdegree, select vertices with largest degrees",
+    //    __FILE__, __LINE__));
+    // </TODO>
+
+    return retval;
+}
+
+/**
+ * @brief Run hello tests
+ * @tparam     GraphT        Type of the graph
+ * @tparam     ValueT        Type of the distances
+ * @param[in]  parameters    Excution parameters
+ * @param[in]  graph         Input graph
+...
+ * @param[in]  target        where to perform the app 
+ * \return cudaError_t error message(s), if any
+ */
+template <typename GraphT>
+cudaError_t RunTests(
+    util::Parameters &parameters,
+    GraphT           &graph,
+    // <TODO> add problem specific reference results, e.g.:
+    typename GraphT::ValueT *ref_degrees,
+    // </TODO>
+    util::Location target)
+{
+    
+    cudaError_t retval = cudaSuccess;
+       
+    typedef typename GraphT::VertexT VertexT;
+    typedef typename GraphT::ValueT  ValueT;
+    typedef typename GraphT::SizeT   SizeT;
+    typedef Problem<GraphT>          ProblemT;
+    typedef Enactor<ProblemT>        EnactorT;
+
+    // CLI parameters
+    bool quiet_mode = parameters.Get<bool>("quiet");
+    int  num_runs   = parameters.Get<int >("num-runs");
+    std::string validation = parameters.Get<std::string>("validation");
+    util::Info info("hello", parameters, graph);
+    
+    util::CpuTimer cpu_timer, total_timer;
+    cpu_timer.Start(); total_timer.Start();
+
+    // <TODO> get problem specific inputs, e.g.:
+    // std::vector<VertexT> srcs = parameters.Get<std::vector<VertexT>>("srcs");
+    // printf("RunTests: %d srcs: src[0]=%d\n", srcs.size(), srcs[0]);
+    // </TODO>
+
+    // <TODO> allocate problem specific host data, e.g.:
+    ValueT *h_degrees = new ValueT[graph.nodes];
+    // </TODO>
+
+    // Allocate problem and enactor on GPU, and initialize them
+    ProblemT problem(parameters);
+    EnactorT enactor;
+    GUARD_CU(problem.Init(graph, target));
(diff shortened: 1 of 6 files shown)


bkj:
cc @neoblizz @sgpyc  If you have a chance to take a look at this, I'd love to hear whether you like this approach or not.

gunrock/app/rw/rw_app.cu
```diff
@@ -0,0 +1,338 @@
...
+
+    // <DONE> add app specific parameters, eg:
+    GUARD_CU(parameters.Use<int>(    
+         "walk-length",   
+         util::REQUIRED_ARGUMENT | util::OPTIONAL_PARAMETER,  
+         10,   
+         "vertex id", 
+         __FILE__, __LINE__));
+    // </DONE>
+
```
neoblizz: I think `seed` for random numbers should be a parameter here, default could be `time(NULL)`, I mention this later on as well.

gunrock/app/rw/rw_enactor.cuh
```diff
@@ -0,0 +1,385 @@
...
+        // <DONE> add problem specific data alias here:
+        auto &walks       = data_slice.walks;
+        auto &rand        = data_slice.rand;
+        auto &walk_length = data_slice.walk_length;
+        auto &gen         = data_slice.gen;
+        // </DONE>
+        
+        curandGenerateUniform(gen, rand.GetPointer(util::DEVICE), graph.nodes);
+        
+        GUARD_CU(frontier.V_Q()->ForAll(
```
neoblizz: Noting that this doesn't do the sorted RandomWalk, which doesn't seem hard because you can use CUB's blocksort to implement that. I think my initial take on it just load balancing nicely might be an incorrect assumption. I'll see if I can find Shari's notes on that.

tests/rw/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+include ../BaseMakefile.mk
+ALGO = rw
+EXEC = bin/test_$(ALGO)_$(NVCC_VERSION)_$(ARCH_SUFFIX)
+
+test: $(EXEC)
+
+$(EXEC) : test_$(ALGO).cu $(DEPS)
+	mkdir -p bin
+	$(NVCC) -w $(DEFINES) $(SM_TARGETS) -o $(EXEC) test_$(ALGO).cu $(EXTRA_SOURCE) $(NVCCFLAGS) -lcurand $(ARCH) $(INC) -O3
```
neoblizz: Add `-lcurand` flag to `NVCCFLAGS`:

somewhere before `test:`
```
NVCCFLAGS += -lcurand
```

neoblizz:
I hope the comments are helpful, @sgpyc could take a look at it as well. :)

gunrock/app/rw/rw_problem.cuh
```diff
@@ -0,0 +1,379 @@
...
+            GUARD_CU(rand.EnsureSize_(nodes, target));
+            // </DONE>
+
+            // Reset data
+            // <DONE> reset problem specific data, e.g.:
+            GUARD_CU(walks.ForEach([]__host__ __device__ (VertexT &x){
+               x = (VertexT)0;
+            }, nodes * this -> walk_length, target, this -> stream));
+            GUARD_CU(rand.ForEach([]__host__ __device__ (float &x){
+               x = 0.0;
```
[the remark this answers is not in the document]
bkj: fixed

gunrock/app/rw/rw_enactor.cuh
```diff
@@ -0,0 +1,385 @@
...
+     * \return cudaError_t error message(s), if any
+     */
+    cudaError_t Enact(
+        // <DONE> problem specific data if necessary, eg
+        // </DONE>
+    )
+    {
+        cudaError_t retval = cudaSuccess;
+        GUARD_CU(this -> Run_Threads(this));
+        util::PrintMsg("GPU Template Done.", this -> flag & Debug);
```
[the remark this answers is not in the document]
bkj: fixed

gunrock/app/rw/rw_app.cu
bkj: fixed

gunrock/app/rw/rw_enactor.cuh
```diff
@@ -0,0 +1,385 @@
...
+        IterationT;
+
+    Problem *problem;
+    IterationT *iterations;
+
+    /**
+     * @brief hello constructor
+     */
+    Enactor() :
+        BaseEnactor("Template"),
```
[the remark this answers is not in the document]
bkj: fixed

gunrock/app/rw/rw_problem.cuh
```diff
@@ -0,0 +1,379 @@
...
+            int walk_length_)
+        {
+            cudaError_t retval  = cudaSuccess;
+
+            GUARD_CU(BaseDataSlice::Init(sub_graph, num_gpus, gpu_idx, target, flag));
+
+            // <DONE> allocate problem specific data here, e.g.:
+            walk_length = walk_length_;
+            curandCreateGenerator(&gen, CURAND_RNG_PSEUDO_DEFAULT);
+            curandSetPseudoRandomGeneratorSeed(gen, time(NULL));
```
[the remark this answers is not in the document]
bkj: fixed

tests/rw/Makefile
bkj: fixed

gunrock/app/rw/rw_enactor.cuh
bkj: OK sounds good -- I didn't look closely at the sorted code.  If you have any references for what Shari was trying to do w/ the sorting I'd love to take a look

bkj:
@neoblizz very helpful -- thanks!

gunrock/app/rw/rw_enactor.cuh
```diff
@@ -0,0 +1,385 @@
...
+        util::Location    target = util::DEVICE)
+    {
+        cudaError_t retval = cudaSuccess;
+        this->problem = &problem;
+
+        // Lazy initialization
+        GUARD_CU(BaseEnactor::Init(
+            problem, Enactor_None,
+            // <OPEN> change to how many frontier queues, and their types
+            2, NULL,
```
sgpyc: RW only uses a single vertex frontier, so should be `1, NULL,` here

gunrock/app/rw/rw_enactor.cuh
sgpyc: Array1D provides the sort() function https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/oprtr/1D_oprtr/sort.cuh#L35 here.

`ForAll` assumes all per-element workload to have ~the same cost, and it doesn't do load balancing. I don't think the random access could cost more than 3 sorting, but I maybe wrong here.

sgpyc:
The RW implementation looks great. Thanks @bkj 

A few things:
1. Removing comments for things ready done?
2. How the 3 TODOs in the check list will show up if we merge now? Or shall they be covered before merging?
 C/Python interface and correctness checking should not be hard. For the uniform distribution part, maybe we can verify by this: get a distribution of relative position of the vertex in its parent's neighbor list. For example, edge <v -> u> is part of a path, then the relative position should be (the position of u in v's neighbor list) / (v's neighbor list length). On large scale, this distribution should be ~uniform.
 Multi-GPU support I don't think can be done so soon, before the refactored framework itself to be tested on multiple GPUs.

bkj:
@sgpyc

Agree that interfaces and correctness checking is easy -- I just wasn't going to bother doing it until you all verified this was a reasonable implementation.  I'll make those changes in the next couple of days.

Anything left in my checklist when we merge can just be moved to issues.

gunrock/app/rw/rw_enactor.cuh
bkj: resolved

gunrock/app/rw/rw_enactor.cuh
neoblizz: I think it will be interesting to get the numbers with sort as well then, because per Shari's code, that performed better than random accesses.

jowens: Sorting potentially increases locality of reference, which leads to better coalescing. It's a win if there is any locality. It is frankly hard to come up with principled ways to improve random walks because one would imagine you would be completely gated by the throughput of random memory accesses, and it's hard to extract any locality from randomness.

neoblizz:
@bkj for correctness checking it is also worth looking at the CPU implementation in place in `dev`, I think there isn't a perfect way to determine and compare randomness generated through `curand` vs. some cpu `rand()` version, but cpu could be a sanity check.

I will also be interested in seeing how sorting would improve locality, I am guessing even though the accesses are random, there's still some locality within a neighborhood of vertices. (just a note here to explore in future)

bkj:
Thanks for the merge -- I found a bug in this implementation this morning, so keep your eyes open for a hotfix patch later today.

In terms of correctness checking -- I think it makes most sense to just implement a function that makes sure that the output is a valid walk starting from the right place.  The [random code is fairly simple](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/app/rw/rw_enactor.cuh#L128) -- if you take a quick look at that, I think we can assume it's correct.

PR #389, opened 2018-09-12
[dev-refactor][HIVE] localgraphclustering/pr_nibble

bkj:
## pr_nibble (localgraphclustering)

This pull request implements PageRank-Nibble, one of the algorithms in the `localgraphclustering` reference implementation on Gitlab.

@ctcyang is also working on an implementation of this app in a GraphBLAS backend, but it was a natural fit for Gunrock and not too difficult to implement.

The Gunrock app is a little more complicated than eg. `sssp` -- each iteration requires some ForAll's in addition to the advance/filter operation.

The `pygunrock` implementation is here:
 - https://github.com/bkj/pygunrock/blob/master/apps/pr_nibble.py

#### TODO
__This shouldn't be merged before the *'d items below are completed__
 - [x] check grad for converged in `pr_nibble_enactor` (*)
 - [x] docs (*)
 - [x] c and python interfaces (*)
 - [ ] accept multiple src nodes (optional?)
 - [ ] multi-gpu implementation


bkj:
cc @ctcyang @sgpyc @neoblizz

bkj:
@ctcyang @neoblizz @neoblizz -- this is ready to be reviewed now.

- added convergence checking
- docs
- external APIs

neoblizz:
I was tagged twice, tagging @sgpyc for review as well. I'll take a look at this tonight.

ctcyang:
I will take a look at this tonight as well.

ctcyang:
I was able to compile your code and run it on the JHU dataset. The result seems to match the CPU reference implementation. One thing that concerns me is that it is based it off the kfoynt implementation, but mine is based off Julian Shun's implementation which will make it a bit hard to compare exact numbers, because even the parameters seem to be different (e.g. Julian's doesn't have rho and phi, but does have alpha and epsilon). Do you know which algorithm the USC people implemented for OpenMP? Maybe we can compare by sorting the PageRank values in descending order after 20 iterations?

I did see a speed-up (13.4ms vs. 103ms from CPU reference) even on the small JHU dataset "./bin/test_pr_nibble_8.0_x86_64 --graph-type market --graph-file jhu --src 0 --max-iter 20 --eps 0.0000001", which is good!

LGTM

bkj:
Great thanks for taking a look! 

Shun and kfoynt's algorithms are two different ways to optimize the same function, so I think we should be able to compare them in theory. We also may be able to translate the parameters -- in kfoynts code, some of the parameters are simple functions of one another.

bkj:
(I'm not sure which algorithm USC implemented -- I asked them but they haven't responded.)

gunrock/app/pr_nibble/pr_nibble_enactor.cuh
```diff
@@ -0,0 +1,539 @@
...
+        ));
+        
+        cudaDeviceSynchronize();
+        cudaMemcpy(data_slice.h_grad_scale, d_grad_scale, 1 * sizeof(int), cudaMemcpyDeviceToHost);
+        
+        GUARD_CU2(cudaDeviceSynchronize(),
+            "cudaDeviceSynchronize failed");
+        
+        return retval;
+    }
```
neoblizz: I am trying to think if the operators you have here can be somehow reduced a bit, currently there is kernel fusion for advance and filter, but operators are serial execution (with I believe a device sync, therefore costly). I don't think I can provide much input on the implementation (and everything else looks pretty straight forward), but it is worth investing some time to see what we can do better with ops.

gunrock/app/pr_nibble/pr_nibble_enactor.cuh
```diff
@@ -0,0 +1,539 @@
...
+        IterationT;
+
+    Problem *problem;
+    IterationT *iterations;
+
+    /**
+     * @brief pr_nibble constructor
+     */
+    Enactor() :
+        BaseEnactor("Template"),
```
neoblizz: `"Name of Template"`

gunrock/app/pr_nibble/pr_nibble_problem.cuh
```diff
@@ -0,0 +1,527 @@
...
+
+    // ----------------------------------------------------------------
+    // Dataslice structure
+
+    /**
+     * @brief Data structure containing problem specific data on indivual GPU.
+     */
+    struct DataSlice : BaseDataSlice
+    {
+        // problem specific storage arrays:
```
neoblizz: Some comments on what these are will be helpful too.

gunrock/app/pr_nibble/pr_nibble_enactor.cuh
bkj: Can you explain what you mean?  I'm not sure I understand

neoblizz: `compute_op` -> `ForAll`
`advance_op` and `filter_op`
convergence_op -> `ForEach`

These three steps are in serial, but within them there should be a lot of parallelism.
ForAll and ForEach though are not load balanced.

It checks for convergence after the advance + filter, so I was wondering if there is a way to represent this algorithm in a way where we can utilize advance and filter more or reduce the loops (because they have a global barrier afterwards, and barriers are costly in CUDA).

Side note: the implementation itself worked perfectly, I was just noting down that there may be room for improvement in terms of performance.

bkj: Yeah interesting.

I don't think we can check for convergence in the exact same way as the original implementation w/o a barrier between `advance_op` and `convergence_op` -- `convergence_op` has to wait for all `grad` to be computed. 

However, we could merge the `compute_op` and the `advance_op` by doing something like:
```
bool src_was_touched = atomicOr(src_touched[src], true);
if(!src_was_touched) compute_op(src);
```
at the top of `advance_op`.  Do you think that might help w/ performance?

neoblizz: Yes, I think so. You now don't have a global barrier, also utilizing load balancing strategies of advance and you have good amount of work in the whole kernel.

bkj: OK cool, I can make that change.

Can you confirm that's a reasonable way to "mark" a node?

neoblizz: Are multiple nodes trying to mark the node valid?

neoblizz: -- I am thinking if all of the nodes are just trying to mark the active node valid, then do you really need to have an atomic there? (another way of asking this question is are there any nodes marking this `false`)?

bkj: Hm this gets a little more complicated actually:  all the other threads w/ the same `src` value have to block until `compute_op(src)` is done.  Is there a way to do that without a global barrier?

neoblizz: Yes, you use `__syncthreads()` for this, where you want all threads to be done with whatever `compute_op` needs to do.

neoblizz: __syncthreads() is cheaper than doing a global sync, but notice that this is a barrier per block.

neoblizz: This may be assuming that the `src` you are working on (all threads working on that) are in a single block. Which could be possible, gunrock a lot of times has a filling launch with fewer blocks (e.g. 2) and lots of threads (e.g. 1024).

neoblizz:
This is ready for merge, just wanna check in if you can somehow condense the operators by doing `compute_op` in the advance? If it is too complicated for now, we can create issues for this and merge the initial implementation in. @bkj

bkj:
It's too complicated for right now I think, I'll take a look at it as an optimization later.

PR #392, opened 2018-09-24
[dev-refactor][HIVE] geolocation/geotagging

neoblizz:
# geolocation

Using two `ForAll()` to do a gather (populate locations) and compute (spatial kernel). I still have a few things to sort out, I moved my initial implementation to the new `hello` template, because its a lot cleaner.

**TODOs - do not merge until the following are completed:**
- [x] Add the preprocessing step for changing user labels (latitude and longitude) to locations
- [x] Complete spatial kernel calculations
- [ ] Add c and python interfaces **(optional)**
- [x] document the application https://github.com/gunrock/docs/pull/2

I will have the spatial kernel done soon, its already wip, just debugging an issue with my calculation for spatial median with > 3 locations. **(done)**

I am also using compressed notation to store latitude and longitude, need better memory utilization than just allocating a large locations array to store info. I think this step can be done better. **(done)**

**NOTE:** I'll request a review once I am done with spatial kernel stuff.


neoblizz:
Talked to Yuechao on memory usage, two ways I can improve this:
- [ ] Declare `Dins[]` array to be `|E|` and fix the indexing for it accordingly.
- [ ] Recompute the valid neighbors' locations instead of storing them before hand (`Get` method). Storing them requires `|V|x|V|` sized array, or smaller with some tricks, but that is not ideal.

neoblizz:
@bkj you can still take a review rest of the stuff if you have time, thanks!

neoblizz:
@jowens understood, I will make appropriate notes in the report regarding this.

neoblizz:
I had some time to work on this after the report, upcoming commit:
- Removed `D[]` and `W[]` arrays as they can be computed on the fly within a register.
- `Dinv[]` inverted distance array is now of size `|E|`. (largest array in the app, if more space needed for larger dataset this can be replaced with redundant computation.
- `LocationsList` is also removed, now recomputed for each spatial center calculations: `mean()` and `spatial_median()`.

bkj:
This looks pretty good to me.   See review for some minor comments,

gunrock/app/geo/geo_app.cu
```diff
@@ -0,0 +1,376 @@
...
+        "Run geolocation application until all locations for all nodes are found, uses an atomic (default=false).",
+        __FILE__, __LINE__));
+
+    GUARD_CU(parameters.Use<std::string>(
+        "labels-file",
+        util::REQUIRED_ARGUMENT | util::SINGLE_VALUE | util::OPTIONAL_PARAMETER,
+        "",
+        "User locations label file for geolocation app.",
+        __FILE__, __LINE__));
+
```
bkj: (low priority) You may want to expose the`max_iter` for `spatial_median` to the user via the command line.  Some of the TA1/2 performers were talking about fixing the number of iterations (to a number ~ 10 instead of ~1000) in order to get an apples-to-apples comparison in the face of floating point errors, etc

gunrock/app/geo/geo_app.cu
```diff
@@ -0,0 +1,376 @@
...
+        GUARD_CU(problem.Reset(
+            // <DONE> problem specific data if necessary, eg:
+            h_latitude,
+	    h_longitude,
+	    geo_iter,
+            // </DONE>
+            target
+        ));
+        GUARD_CU(enactor.Reset(
+            // <TODO> problem specific data if necessary:
```
bkj: Probably a good idea to remove complete `<TODO>`'s, so people looking at the code don't think that this is an incomplete implementation.

gunrock/app/geo/geo_d_spatial.cuh
```diff
@@ -0,0 +1,354 @@
+// ----------------------------------------------------------------
```
bkj: (low priority) Is it possible to combine this file w/ `geo_spatial.cuh`?  (That's an actual question, not a rhetorical question).  Would reduce the amount of code significantly, esp. since these spatial helper functions aren't super obvious

gunrock/app/geo/geo_enactor.cuh
```diff
@@ -0,0 +1,490 @@
...
+				latitude,
+				longitude,			
+				v,
+				D, Dinv, W,
+				false);
+
+	    } // -- median calculation.
+	};
+
+	auto status_op = [
```
bkj: I think this could be done inside `gather_op` or `compute_op`, maybe saving an extra pass over the frontier

gunrock/app/geo/geo_d_spatial.cuh
bkj: Also, I didn't check every line of the math here, since you said you're confident in the correctness testing

gunrock/app/geo/geo_test.cuh
```diff
@@ -0,0 +1,219 @@
...
+template <typename GraphT>
+typename GraphT::SizeT Validate_Results(
+             util::Parameters &parameters,
+             GraphT           &graph,
+             typename GraphT::ValueT *h_predicted_lat,
+             typename GraphT::ValueT *h_predicted_lon,
+             typename GraphT::ValueT *ref_predicted_lat,
+             typename GraphT::ValueT *ref_predicted_lon,
+             bool verbose = true)
+{
```
bkj: No actual correctness checking is being done here.  Presumably you'll need to check within some epsilon difference, since spatial median might converge to slightly different solutions depending on eg. order of the data

gunrock/app/geo/geo_d_spatial.cuh
neoblizz: Yes, I think I can combine the two files if I use `Array1D` representation for host arrays as well. I'll try to merge it together to reduce code.

gunrock/app/geo/geo_app.cu
neoblizz: Done.

gunrock/app/geo/geo_app.cu
neoblizz: Done. (maybe I missed some)

gunrock/app/geo/geo_enactor.cuh
neoblizz: I don't use this operator anyways, it is just if someone wants to do a complete geo run. Not used for performance measurement. Keeping it separate so it looks clean.

gunrock/app/geo/geo_d_spatial.cuh
neoblizz: Done!

neoblizz:
- [x] Formatting done.
- [x] All comments addressed.

PR #393, opened 2018-09-25
[dev-refactor][HIVE] bkj/graph_search

bkj:
### GraphSearch

`GraphSearch` is a small modification to random walks -- each node has a score, and the walk chooses a neighbor based on some function of the neighbor scores:
 - uniform: ignores scores, yielding a standard random walk
 - max: walks to neighbor w/ max score
 - biased random sampling: sample proportional to (softmax?) of neighbor scores

`rw` already implemented the uniform sampling method; this PR implements the max sampling method, plus the code to load node features from a file.  It should be pretty trivial to add other sampling methods, but I think they should really be done w/ a neighborhood reduction instead of the current serial method, so we can hold off on that for now.

### TODO:
 - [ ] Implement biased random sampling (optional)

### DIFF
--- a/gunrock/app/rw/rw_enactor.cuh
+++ b/gunrock/app/rw/rw_enactor.cuh
@@ -54,7 +54,7 @@ struct RWIterationLoop : public IterationLoopBase
     typedef typename EnactorT::ValueT  ValueT;
     typedef typename EnactorT::Problem::GraphT::CsrT CsrT;
     typedef typename EnactorT::Problem::GraphT::GpT  GpT;
-    
+
     typedef IterationLoopBase
         <EnactorT, Use_FullQ | Push> BaseIterationLoop;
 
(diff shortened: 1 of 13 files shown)


neoblizz:
I think instead of neighbor reduce (for now) we can actually use a `ForAll` with `for` to do this pretty easily, thing is neighbor reduce's load balancing might not be the right solution for all types of graphs, and you still get parallelism with `ForAll`, just that you have more work inside the operator.

So, graphs that have smaller max degrees (like road networks) will do better with decent work and parallelism on `ForAll()`, instead of load balancing with neighbor reduce.

bkj:
@neoblizz If I understand you correctly, I think `ForAll` w/ `for` is what it's doing now -- 
https://github.com/bkj/gunrock/blob/bkj/graph_search/gunrock/app/rw/rw_enactor.cuh#L141

gunrock/app/rw/rw_enactor.cuh
```diff
@@ -69,50 +69,91 @@ struct RWIterationLoop : public IterationLoopBase
...
+                VertexT neighbor     = graph.GetEdgeDest(neighbor_list_offset + offset);
+                ValueT  neighbor_val = graph.node_values[neighbor];
+                if(neighbor_val > max_neighbor_val) {
+                  max_neighbor_id  = neighbor;
+                  max_neighbor_val = neighbor_val;
+                }
+              }
+              v[i] = max_neighbor_id; // Replace vertex w/ neighbor in queue
+            }
+          };
```
neoblizz: Just a note here, is it better to have two separate lambdas for the different random walk modes, or have the if statement just inside one single lambda?

-- on a performance note, the predictor will learn very quickly that the walk mode is constant and the if statement won't diverge.

Let me know your thoughts on this (what's the best programming practice here), if statement inside lambda to select the mode, or outside?

bkj: I would prefer using two separate lambdas, as it is right now -- that way, someone could reasonably write a new lambda without having to touch the existing ones.

PR #395, opened 2018-09-26
Generates graph as an edge list using grmat

rgsl888prabhu:
There was no explicit option to use grmat functionality to generate the graph for external use. 

This implementation can write the edge list information to a file or return a pointer which can be used. For pointer usage, the user has to modify the function as currently the pointer is getting freed at the end.

I am working in NVIDIA and this implementation would be helpful in different scenarios.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,10 @@ option(GUNROCK_APP_TOPK
   "If on, builds only TOPK application."
   OFF)
 
+option(GUNROCK_APP_GRMAT
+  "If on, builds only GRMAT application."
+  OFF)
+
 #option(GUNROCK_APP_SAMPLE
 #  "If on, builds only SAMPLE application."
 #  OFF)
@@ -294,6 +298,7 @@ if(GUNROCK_BUILD_APPLICATIONS)
   add_subdirectory(tests/salsa)
   add_subdirectory(tests/wtf)
   add_subdirectory(tests/topk)
+  add_subdirectory(tests/grmat)
   #add_subdirectory(tests/template)
   #add_subdirectory(tests/vis)
   #add_subdirectory(tests/mis)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -432,6 +445,10 @@ if(GUNROCK_MGPU_TESTS)
   add_test(NAME TEST_MGPU_PR COMMAND pr market
     ${gunrock_INCLUDE_DIRS}/simple_example/bips98_606.mtx ${DEVICES} --undirected)
   set_tests_properties(TEST_PAGERANK PROPERTIES FAIL_REGULAR_EXPRESSION "INCORRECT")
+  
+  add_test(NAME TEST_MGPU_GRMAT COMMAND grmat market
+    ${gunrock_INCLUDE_DIRS}/simple_example/bips98_606.mtx ${DEVICES} --undirected)
+  set_tests_properties(TEST_GRMAT PROPERTIES FAIL_REGULAR_EXPRESSION "INCORRECT")
```
neoblizz: We don't need GRMAT `ctests`, and there is no shared library implementation for it either. So, please remove this.

CMakeLists.txt
```diff
@@ -371,6 +380,10 @@ add_test(NAME TEST_TOPK COMMAND topk market
   ${gunrock_INCLUDE_DIRS}/simple_example/bips98_606.mtx --undirected)
 set_tests_properties(TEST_TOPK PROPERTIES FAIL_REGULAR_EXPRESSION "INCORRECT")
 
+add_test(NAME TEST_GRMAT COMMAND grmat market
+  ${gunrock_INCLUDE_DIRS}/simple_example/bips98_606.mtx --normalized --compensate --undirected)
+set_tests_properties(TEST_GRMAT PROPERTIES FAIL_REGULAR_EXPRESSION "INCORRECT")
```
neoblizz: See next comment, we don't need GRMAT `ctests`, also, I don't think the app you have outputs the expression "CORRECT" or "INCORRECT", so I am not sure if this will ever fail.

CMakeLists.txt
rgsl888prabhu: I will make these changes, I had kept it just so that in future if we plan to add anything, it would be easier. But this can be done easily, I will remove it.

CMakeLists.txt
rgsl888prabhu: I will remove this.

CMakeLists.txt
neoblizz: Yeah, should be pretty easy, but we might end up using googletest instead of ctest for these kind of cases.

rgsl888prabhu: I have made the changes and committed, please review

PR #399, opened 2018-09-28
[dev-refactor][HIVE] graph_projections

bkj:
## Graph Projections

Given a directed graph `G`, `GraphProjections` outputs a graph `Gp` s.t. there is an edge `(u,v)` in `Gp` iff there are edges `(t,u)` and `(t,v)` in `G`.

#### Reference implementations
 - (C++) https://gitlab.hiveprogram.com/jfiroz/graph_projection
 - (pygunrock) https://github.com/bkj/pygunrock/blob/master/apps/graph_projections.py

#### Issues
 - It's a little unclear what the intended use case/topology for this is.  The HIVE slides suggest it's a bipartite graph w/ node sets `U` and `V` s.t. `|V| >> |U|`.  However, the reference implementation is designed to work for arbitrary graphs.  The C++ version above has a flag for bipartite vs. unipartite graphs, but I'm not sure it behaves as desired (I have a question out to the author).
 - Often these projected graphs get _very_ dense.  With large graphs, there's a risk that we run out of GPU memory: graph projections can increase the number of edges in a graph by > 4 orders of magnitude (!)  How should we deal with this?
 - Right now, we're allocating a dense `|V|x|V|` array to store the projected edges -- this is clearly not memory efficient, and should be replaced if we want to support large graphs.

#### TODO
 - [x]  `proj` should _only_ take directed graphs -- how do we set that to be default? (cc @sgpyc)
 - [ ]  (optional) Implement additional weighting functions

[1] https://arxiv.org/pdf/1611.02756.pdf

### DIFF
--- a/gunrock/app/proj/proj_app.cu
+++ b/gunrock/app/proj/proj_app.cu
@@ -1,0 +1,337 @@
+// ----------------------------------------------------------------------------
+// Gunrock -- Fast and Efficient GPU Graph Library
+// ----------------------------------------------------------------------------
+// This source code is distributed under the terms of LICENSE.TXT
+// in the root directory of this source distribution.
+// ----------------------------------------------------------------------------
+
+/**
+ * @file proj_app.cu
+ *
+ * @brief Simple Gunrock Application
+ */
+
+#include <gunrock/gunrock.h>
+#include <gunrock/util/test_utils.cuh>
+#include <gunrock/graphio/graphio.cuh>
+#include <gunrock/app/app_base.cuh>
+#include <gunrock/app/test_base.cuh>
+
+#include <gunrock/app/proj/proj_enactor.cuh>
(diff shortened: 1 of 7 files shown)


neoblizz:
> How do you force `directed` graphs usage only?

@sgpyc commented:
> Don't give the 'util::UNDIRECTED' flag to type switch, and force 'undirected' parameter to be false
> But if the .mtx says it's symmetrical, there's nothing much can be done: it's a wrong input

Which I think you already have in place in `test_proj.cu` file.

bkj:
What do you mean it's a wrong input?  The flags in the `.mtx` file override the command line flags?

neoblizz:
I think so, mtx's header overrides the flags.

neoblizz:
Merging for now, I don't see a problem with this PR. Optional stuff could be another PR.

PR #403, opened 2018-10-07
[New operator] neighbor reduce

sgpyc:
Should merge after the Louvain PR
Provides a working neighbor reduce implementation, tested with pulled direction PageRank.
Performance not optimized yet


neoblizz:
I think what will be really useful for all of us is a stripped down usage of the API (aside from all the PageRank stuff in place). Thank you for implementing this, I will try using this in my refactor for `color`.

PR #407, opened 2018-10-09
[Louvain] Performance improvements using load balanced segmented reduce and avoiding segmented sort

sgpyc:
Mergeable after the neighbor reduce operator and MF
`--unify-segments` is false by default, need to switch to true after more testing


neoblizz:
Conflicts resolved, merging this in. Try the rebase now for future pull-requests. Thanks for cleaning this up!

PR #412, opened 2018-10-15
[graphio/market] loader using binary storage

sgpyc:
The market loader will try to read a binary copy of the graph, before the actual .mtx file. If binary copy does not exit, it will try to store a copy after reading the .mtx file.
Only one copy of binary will be stored, regardless of options (undirected, 64bit-SizeT, 64bit-VertexT, etc.).
Graph modifications (edge doubling, subtract 1 from vertex Ids, removing self loops, store to binary, etc.) will be print out
To reviewer: the commit history is still messed up, the last 4 commits are the actual PR. Files changed log seems accurate; the main changes are in the graph or graphio related files


neoblizz:
Didn't see anything out of ordinary in this PR. Merging the PR instead of squash, maybe that will fix the duplicated commits.

PR #413, opened 2018-10-16
[HIVE][do-not-merge] bkj/testing-hotfix

bkj:
This is a couple of minor changes to improve OOM handling in the `rw` and `proj` apps.

__Note:__ I'm not sure why the git history looks like this -- the `Files Changed` tab shows this is just a small number of changes.  I think it's because this is based off of @sgpyc 's `dev-refactor`, so the commits show up individually instead of as merges.  AFAICT this should be fine to merge.


bkj:
There might be a bug here -- hold off on merging until I track it down.

bkj:
I'm closing this because the I manually moved the changes to other branches to avoid these git conflicts.  Will reopen if I have to.

PR #416, opened 2018-10-19
Fix build on Ubuntu 18.04.1 LTS

porumbes:
@neoblizz 

The ` -std=c++11` parameter was being passed to nvcc multiple times and gunrock would not build.

`test_sm.cu` was passing an `unsigned long long` to the abs function. This is ambiguous because abs expects a signed value.


neoblizz:
I was thinking if we move the `set(CUDA_NVCC_FLAGS -std=c++11)` line before `googletest` is added, it may solve the problem with the c++11 flag added to nvcc instead of cxx. I am merging it to test the build for now.

PR #419, opened 2018-10-23
[HIVE] bkj/test_hotfix

bkj:
These are a bunch of relatively small changes to `vn`, `rw` and `proj` that I needed to make for HIVE testing.  Couple of bugfixes, couple of API changes.

Should have no conflicts w/ `dev-refactor`.

~ Ben


neoblizz:
I don't see any obvious thing that's wrong with all of these commits. Were you able to verify that adding a synchronize after curandgenerator fixed the problem of async generation of states?

bkj:
@neoblizz I put it on the same stream as Gunrock, so it should be fixed.  Also, the screenshots from the profiler I sent you earlier were the result of a bug -- in reality, the advance takes >> longer than the RNG, so the overlap between the two kernels was way overstated in those pictures.

PR #466, opened 2019-01-23
Apply "Fix pr (PageRank) build error on dev-refactor." to dev branch #461

porumbes:
Bring the SM architecture check for atomic with doubles over from dev-refactor.


neoblizz:
This is fine, the upstream isn't that big of an issue.

PR #469, opened 2019-01-26
Color Optimizations: atomic fix, removed syncs

neoblizz:


Samtruong:
I think all the changes make sense. I will just summarize what i find that is new, so that we can both double check.

color_app.cuh:
- Change parameters to info.SetVal to fix the num-colors default problem. There are unnecessary flags like min-color, loop-color that can be erased.

color_enactor.cuh:
- Regenerate rand array every other iteration, helps large graphs to converge because of same rand issue.
- RepeatFor  segment is disabled for luigi CUARCH=3.5
- Reset colored inside jpl_color_op and color_op, optimize atomic count
- color_op does not have iteration % 2 rand regeneration. Result have shown that it can achieve 1-2 colors less with regeneration (maybe we can add it? I have the regeneration on my forked)
- Atomic add does not need reset anymore

PR #478, opened 2019-02-02
Counting clusters and core-points for json.

neoblizz:


neoblizz:
@agalup this is how I changed it.

Issue #481, opened 2019-02-04
rgg row_offsets bug


maawad:
**Summary:**
Generated RGG graph `row_offsets`'s last two values are not correct. `row_offsets[i+1] < row_offsets[i]` for the last vertex.

**Reproducibility:**
Always using the steps below.

**To Reproduce:**
In dev-refactor branch. You can add this debugging code in [`rgg.cuh`](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/graphio/rgg.cuh#L380) after the graph is generated. 
```
for(VertexT v = 0; v < num_nodes; v++)
{
	SizeT end =  graph.CsrT::row_offsets[v+1];
	SizeT start = graph.CsrT::row_offsets[v];
	if(start > end)
	{
		printf("RGG build error at V[%u]: "
			"row_offsets[v+1] = %u,"
			"row_offsets[v] = %u.\n", v, end, start);
	}
}
```
Then run sssp with `rgg` argument.
`./bin/test_sssp rgg`

Debug result:
`RGG build error at V[1023]: row_offsets[v+1] = 0,row_offsets[v] = 2732`

**Actual Results:**
SSSP runs without errors.
Probably related to #458. Setting `row_offsets[num_nodes]` to  `num_edges` manually makes that problem disappear.

**System Configuration:**
Luigi.
GPU: K40c
CPU: Intel(R) Xeon(R) CPU E5-2637 v2 @ 3.50GHz
OS: Ubuntu 16.04.5 LTS
NVIDIA-SMI 410.72
Driver Version: 410.72
CUDA Version: 10.0
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609

maawad:
This last loop ([rgg.cuh](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/graphio/rgg.cuh#L350)) never touches the last entry in `row_offsets`. So, I think the easy solution is just to add this statement in the end (after the parallel work is done):
`row_offsets[num_nodes] = row_offsets[num_nodes-1]`

maawad:
This doesn't look correct as well ([rgg.cuh](https://github.com/gunrock/gunrock/blob/dev-refactor/gunrock/graphio/rgg.cuh#L297)). If the point lies inside the first two rows (or columns) this loop will never execute.

Something like this is one fix:
```
SizeT x_start = x_index < 2 ? 0 : x_index - 2;
[...]

PR #485, opened 2019-02-15
RGG: updating row_offsets last value and fix grid boundary bug

maawad:
#481 
* Fix last entry in `row_offsets`.
* Handle boundary scenarios for first two rows/columns of the grid.


neoblizz:
Testing this on my end, then merging.

Issue #459, opened 2019-01-16
Code-coverage should be optional


neoblizz:
If this is enabled, it will require installation of `lcov`, so we should only enable this if `GUNROCK_CODE_COVERAGE` variable is set, just like how we do it for `GUNROCK_GOOGLE_TESTS`:

https://github.com/gunrock/gunrock/blob/fd9e1ce6c5f644b3c9478108727b40defc9dbf00/CMakeLists.txt#L89-L93

PR #486, opened 2019-02-17
Codecov is now optional

neoblizz:
+ Minor change: Moved `TC` from global_indicator to app.


neoblizz:
Solves issue #459

neoblizz:
I just moved `TC` into a proper directory, it was chilling alone in `global_indicator`. Didn't really fix anything there. This PR is just "Codecov is now optional."

Issue #503, opened 2019-03-01
WTF: Illegal Memory Access


neoblizz:
**Summary:**
Running WTF application results in illegal memory access.

**Reproducibility:**
Always

**To Reproduce:**
```
cd examples/wtf
make
./bin/test_wtf_10.0_x86_64 market ../../dataset/small/chesapeake.mtx
```

**Expected Behavior:**
I am not sure what the expected output is, something like (but this result is buggy):
```
GPU result.Top 10 Page Ranks:
Vertex ID: -1728455816, Page Rank: -0.000000
Vertex ID: 32735, Page Rank: 0.000000
Vertex ID: -1728455816, Page Rank: -0.000000
Vertex ID: 32735, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
Vertex ID: 0, Page Rank: 0.000000
```

**Actual Results:**
```
mosama@mario:~/forks/neo/gunrock/examples/wtf$ ./bin/test_wtf_10.0_x86_64 market ../../dataset/small/chesapeake.mtx
Loading Matrix-market coordinate-formatted graph ...
  Reading directly from stored binary CSR arrays from ../../dataset/small/.chesapeake.mtx.di.0.bin ...
Done reading (0s).

Degree Histogram (39 vertices, 340 edges):
    Degree   0: 0 (0.00%)
    Degree 2^0: 0 (0.00%)
    Degree 2^1: 1 (2.56%)
    Degree 2^2: 22 (56.41%)
    Degree 2^3: 13 (33.33%)
    Degree 2^4: 2 (5.13%)
    Degree 2^5: 1 (2.56%)

Using 1 GPU: [ 0 ].
[../../gunrock/oprtr/advance/kernel.cuh, 411 @ gpu 0] cudaStreamSynchronize failed (CUDA error 77: an illegal memory access was encountered)
[../../gunrock/util/array_utils.cuh, 503 @ gpu 0] counters cudaMemcpyAsync H2D failed (CUDA error 77: an illegal memory access was encountered)
[../../gunrock/util/cta_work_progress.cuh, 347 @ gpu 0] CtaWorkProgress GetQueueLength failed (CUDA error 77: an illegal memory access was encountered)
[...]

PR #513, opened 2019-03-05
Fixes illegal memory access bug in WTF.

neoblizz:
Fix for issue #503, needs further testing to make sure the primitive works as intended.


neoblizz:
Also added a test on `chesapeake.mtx` dataset that passes. I am not super familiar with this algorithm (and haven't read the paper), so in future we can revisit this and make sure the result is correct.

PR #521, opened 2019-03-13
[CSR] adding a sort function

maawad:
Added a function to sort CSR edges per vertex. This is required for current TC implementation. Let me know if something else needs to be added or changed.

### DIFF
--- a/gunrock/graph/csr.cuh
+++ b/gunrock/graph/csr.cuh
@@ -392,6 +392,49 @@ struct Csr :
         return retval;
     }
 
+    /**
+     * @brief Sort CSR graph edges per vertex in ascending order
+     *
+     */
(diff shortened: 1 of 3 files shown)


jowens:
I don't have a strong feeling on how to do this, but will programmers be annoyed if they send in a sorted CSR and then also have to pass `--sort-csr`? More programmer-friendly is to throw the error only if they (don't have a sorted CSR) AND (don't pass `--sort-csr`). This would require an `isCSRSorted` function, presumably.

neoblizz:
If this argument is required for an implementation, can be made required in the `USE_PARAMETERS` interface for just that application/operator instead of making it globally required. We don't need isCSRSorted, just require that flag for TC (or the underlying operator that needs it: intersection?)

PR #529, opened 2019-04-30
Pre Release, merge to dev before the final merge release.

neoblizz:
- Removing dependencies installation scripts, Boost and Metis are now optional.
- Removing old sample and simple_example files.
- Added support for HAS_CSC and HAS_CSR to work together.
- Updating moderngpu's commit to support CUDA 9+.
- Color, Louvain, BC, SSSP, PR c-interface and support.
- CMakeLists.txt build support.
- Shared Library build turned off due to a bug.


neoblizz:
- GTF and MF need more work to be supported in the CMake builds.
- BC, Color, Geo, HITS, KNN, Lovain, PR, LGC (PR_NIBBLE), Proj, RW, SAGE, SS, SSSP, TC, VN are all supported.
- Removed SNN from within the KNN app, needs to be a separate primitive in future release.

PR #555, opened 2019-06-12
Add subgraph matching to pre-release

Laurawly:
@neoblizz could you review?


examples/sm/sm_main.cu
```diff
@@ -48,25 +48,30 @@ struct main_struct
...
+        GraphT data_graph;
+        GraphT pattern_graph;
 
         cpu_timer.Start();
-        GUARD_CU(graphio::LoadGraph(parameters, graph));
+        GUARD_CU(graphio::LoadGraph(parameters, data_graph));
+        GUARD_CU(graphio::LoadGraph(parameters, pattern_graph, "pattern-"));
         cpu_timer.Stop();
+        std::cout << "data_graph: " << data_graph.nodes << " " << data_graph.edges << std::endl;
+        std::cout << "query_graph: " << pattern_graph.nodes << " " << pattern_graph.edges <<std::endl;
```
neoblizz: Can you please use `util::PrintMsg("data_graph: " + std::to_string(data_graph.nodes) ... , !quiet);` instead, so when someone uses the `quiet` flag, this isn't printed out?

examples/sm/sm_main.cu
```diff
@@ -77,16 +82,17 @@ struct main_struct
...
             {
-                return app::sm::RunTests(parameters, graph, ref_subgraph_match);
+                return app::sm::RunTests(parameters, data_graph, pattern_graph, ref_subgraph_match, util::DEVICE);
             }));
 
         if (ref_subgraph_match != NULL)
         {
             delete[] ref_subgraph_match; ref_subgraph_match = NULL;
         }
+        GUARD_CU(pattern_graph.Release());
```
neoblizz: Release the `data_graph` as well.

gunrock/app/sm/sm_enactor.cuh
```diff
@@ -69,44 +70,174 @@ struct SMIterationLoop : public IterationLoopBase
...
+            }
+            return false;
+        };
+        auto filter_op = [subgraphs, nodes_query] __host__ __device__(
+            const VertexT &src, VertexT &dest, const SizeT &edge_id,
+            const VertexT &input_item, const SizeT &input_pos,
+            SizeT &output_pos) -> bool
+        {
+            if (!util::isValid(dest)) {
+                printf("False @ %u\n", dest);
```
neoblizz: Remove printfs from the lambdas.

gunrock/app/sm/sm_enactor.cuh
```diff
@@ -69,44 +70,174 @@ struct SMIterationLoop : public IterationLoopBase
...
-
-	// // Sort the subgraph matching values for each node in descending order
- //        GUARD_CU(util::cubSortPairs(
- //            cub_temp_space,
- //            subgraphs, tats1,
- //            nodes,      nodes1,
- //            graph.nodes, 0, sizeof(ValueT) * 8, stream));
+        frontier.queue_reset = true;
+        size_t pointer_head = 0;
+        int num_init = 3; // for current tests, 3 is enough
```
neoblizz: How is this determined? It needs to be a command line parameter if given a different dataset we need to update this, won't make sense to recompile every time.

gunrock/app/sm/sm_enactor.cuh
```diff
@@ -69,44 +70,174 @@ struct SMIterationLoop : public IterationLoopBase
...
-            graph.csr(), frontier.V_Q(), frontier.Next_V_Q(), 
-            oprtr_parameters, intersect_op));
-
-	// // Sort the subgraph matching values for each node in descending order
- //        GUARD_CU(util::cubSortPairs(
- //            cub_temp_space,
- //            subgraphs, tats1,
- //            nodes,      nodes1,
- //            graph.nodes, 0, sizeof(ValueT) * 8, stream));
+        frontier.queue_reset = true;
```
neoblizz: This will reset the queue length to graph.edges every iteration, even if filter was done. I just wanna note that to check with you if that is what you intended.

Laurawly: Yeah, because I only run enactor once. This is intended.

neoblizz: Ok.

gunrock/app/sm/sm_enactor.cuh
neoblizz: Me:
> How big can the query graph get?
Leyuan:
> It's fairly small.
> Usually 3-6 nodes.

jowens: 3-6 nodes is pretty small. What happens if the query graph gets bigger? Does it scale well? Poorly? Or break? How big are the query graphs in the literature / in GraphChallenge?

PR #567, opened 2019-07-03
fix the permission of the scripts and rename the bin of the test scripts

Z-Y00:
the ./bin/test_bc_9.1_x86_64 will only be generated from cuda 9.1
I changed it to ./bin/test_bc* for compatibility.


yzhwang:
Seems to be a reasonable PR. Thank you @Z-Y00 !! I'm actually not sure if we should also add a script to change permissions for test scripts or just making them executable before commit (like this PR does).

Issue #454, opened 2019-01-11
Get Edge from Source and Destination


neoblizz:
We have get src and dest from an edge, but I want the opposite:

https://github.com/gunrock/gunrock/blob/dev/gunrock/graph/csr.cuh#L465-L470

neoblizz:
Moving this out of release, I will let someone new to gunrock take a shot at this. It is quite simple.

PR #569, opened 2019-07-15
solution to git issue #454

DanLoran:
I have a a debug version of the changed csr.cuh file on my local machine that can be run in nsight debugger to confirm that my function works.

### DIFF
--- a/gunrock/graph/csr.cuh
+++ b/gunrock/graph/csr.cuh
@@ -323,6 +323,7 @@ struct Csr :
         //                + ") != v " + std::to_string(v));
         //    }
         //}
+
         return retval;
     }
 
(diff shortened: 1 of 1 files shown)


porumbes:
Ditto on the binary search. The index information for CSR should be sorted by construction.

@neoblizz @sgpyc @yzhwang Do we have a good example of where this new method could/should be used? It would be nice for a certain level of testing. Either way, I think it would be good to have a little google test of this method (and eventually more).

neoblizz:
> Ditto on the binary search. The index information for CSR should be sorted by construction.
> 
> @neoblizz @sgpyc @yzhwang Do we have a good example of where this new method could/should be used? It would be nice for a certain level of testing. Either way, I think it would be good to have a little google test of this method (and eventually more).

The reason, why I made that issue, was 1) due to the completion of the API and 2) SNN did this by hand, so (SNN or KNN) would be a great example to substitute this code into instead of hardcoding an implementation.

DanLoran:
Committed requested function using BinarySearch. Had some issues using the BinarySearch_RightMost utility that similar functions (GetEdgeSrc, GetEdgeSrcDest) use, am looking into developing a googletest with @porumbes to pinpoint the issue. As it is, this commit currently works by using a different binary search utility.

porumbes:
Looks good with the binary search, but please remove the commented out code.

PR #576, opened 2019-07-29
Streamlining performance testing for gunrock and json fixes

neoblizz:
- v0.5.x scripts are incompatible with the current version and will require switching to the v0.5.1 tag for those to work.
- new scripts can be found in v1.x.x (wip)

    - [ ]  bfs
    - [ ]  sssp
    - [ ]  bc
    - [ ]  cc (needs the port)
    - [ ]  pr
    - [ ]  tc
    - [ ]  ... add more.

- add instructions on how to write scripts that span a good number of datasets and does relevant performance testing


gunrock/util/info_rapidjson.cuh
```diff
@@ -343,11 +356,35 @@ struct Info {
 
   template <typename T>
   void SetVal(std::string name, const std::vector<T> &vec) {
+    if (json_writer == NULL) return;
+    if (vec.size() == 1) {
```
jowens: I'm a little nervous about this. I think it's a legit thing to have a 1-length array. The request I had was simply that a particular array (load balancing type), which was _always_ 1-length, be collapsed. But it would be more complicated to parse if sometimes a particular field was an array and sometimes it wasn't. I'd rather that load-balancing be special-cased to collapse it but otherwise I think you can remove this code, or make it a separate `collapseArray` call and call it on load-balancing.

neoblizz: The reason it happens is because of our new support for `--parameter=option1,option2...`, so it is not just `advance-mode` that would make a one-element array, but also some other fields as well, which will never be an array otherwise.

Do you still think I should add a specific case for `advance-mode`? I would also like a use-case where we'd need a 1-length array and not just have it as a field?

jowens: Why would `option1,option2` generate an array? Shouldn't it always just generate a non-array `parameter = option1` or `parameter = option2`?

jowens: If I ran only one test, then I would expect `srcs` or `process-times` would be a one-element array. From the point of view of parsing, I want the same datatype for any particular field, always. For `srcs`, I would expect an array, no matter what.

neoblizz: > Why would `option1,option2` generate an array? Shouldn't it always just generate a non-array `parameter = option1` or `parameter = option2`?

If it is a multi-input option, it is stored in a vector, and that vector gets translated into an array in RapidJSON.

neoblizz: > If I ran only one test, then I would expect `srcs` or `process-times` would be a one-element array. From the point of view of parsing, I want the same datatype for any particular field, always. For `srcs`, I would expect an array, no matter what.

Ok, let me investigate a bit more tonight, the easiest thing to do would be to just make the`advance-mode` a special case of one field and revert this.

jowens: Right, but each individual JSON run should use only one of the multi-input options. At some point you have to pick out which option and perhaps at that point, that option ought to be written (as a scalar) into the JSON?

jowens: I think every multi-input option that results in only one option in a particular run (in a particular JSON output) should have scalars, not one-element vectors, for that option.

PR #591, opened 2019-09-04
move sm_app header to gunrock.h

Laurawly:
@neoblizz Could you review?

### DIFF
--- a/gunrock/app/sm/sm_app.cu
+++ b/gunrock/app/sm/sm_app.cu
@@ -197,10 +197,7 @@ double gunrock_sm(
  * @param[in]  col_indices CSR-formatted graph input column indices
  * @param[in]  edge_values CSR-formatted graph input edge weights
  * @param[in]  num_runs    Number of runs to perform SM
- * @param[in]  sources     Sources to begin traverse, one for each run
- * @param[in]  mark_preds  Whether to output predecessor info
- * @param[out] distances   Return shortest distance to source per vertex
- * @param[out] preds       Return predecessors of each vertex
+ * @param[out] subgraphs   Return number of subgraphs
  * \return     double      Return accumulated elapsed times for all runs
  */
 template <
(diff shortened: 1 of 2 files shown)


gunrock/app/sm/sm_app.cu
```diff
@@ -245,10 +242,36 @@ double sm(
...
+double sm(
+    const int            num_nodes,
+    const int            num_edges,
+    const int           *row_offsets,
+    const int           *col_indices,
+    const unsigned long *edge_values,
+    const int            num_runs,
+          int           *subgraphs)
+{
+    return sm(num_nodes, num_edges, row_offsets, col_indices,
```
jowens: why have `num_runs` as an argument if it's hardcoded to 1?

Laurawly: I think here is only a default value. @neoblizz correct me if I'm wrong. Other apps do so as well.

Laurawly:
I don't see the error message of codecov ci check.

neoblizz:
> I don't see the error message of codecov ci check.

You can ignore the codecov error, it is saying that you added some code that doesn't get coverage during testing. But that is understandable because we don't have a ctest for c-interface.

Laurawly:
> > I don't see the error message of codecov ci check.
> 
> You can ignore the codecov error, it is saying that you added some code that doesn't get coverage during testing. But that is understandable because we don't have a ctest for c-interface.

I see. So are we good in merging it?

neoblizz:
> > > I don't see the error message of codecov ci check.
> > 
> > 
> > You can ignore the codecov error, it is saying that you added some code that doesn't get coverage during testing. But that is understandable because we don't have a ctest for c-interface.
> 
> I see. So are we good in merging it?

Yes, but let's take care of the template/non-template issue as well before we merge this.

Laurawly:
@neoblizz I think we should do a separate or for that some other apps needs to do the same thing in order to make this consistent

PR #595, opened 2019-09-09
Create app template header for subgraph matching

Laurawly:
@aschaffer Does this pr illustrate your suggestions? @neoblizz Could you review? If this is the way to separate template functions to a header file, I'll update bc as well.

PR #606, opened 2019-09-17
Support for GPU RMAT generator.

neoblizz:
@aschaffer I ended up adding grmat functionality to rmat with a parameter that you can set. Will that work for you, or do you guys need it in a separate file (I wanted to avoid redundant code).

### DIFF
--- a/gunrock/app/hello/hello_test.cuh
+++ b/gunrock/app/hello/hello_test.cuh
@@ -83,7 +83,9 @@ typename GraphT::SizeT Validate_Results(util::Parameters &parameters,
 
   // <TODO> result validation and display
   for (SizeT v = 0; v < graph.nodes; ++v) {
-    printf("%d %f %f\n", v, h_degrees[v], ref_degrees[v]);
+    util::PrintMsg(std::to_string(v) + "\t" + 
+                   std::to_string(h_degrees[v]) + "\t" + 
+                   std::to_string(ref_degrees[v]), !quiet);
   }
   // </TODO>
 
--- a/gunrock/graphio/rmat.cuh
+++ b/gunrock/graphio/rmat.cuh
@@ -179,7 +179,19 @@ cudaError_t UseParameters(util::Parameters &parameters,
       0.05, "d for rmat generator, default is 1 - a - b - c", __FILE__,
       __LINE__));
 
+<<<<<<< Updated upstream
   return retval;
+=======
+    GUARD_CU(parameters.Use<bool>(
+        graph_prefix + "grmat",
+        util::REQUIRED_ARGUMENT | util::SINGLE_VALUE | util::OPTIONAL_PARAMETER,
+        false,
+        "Generate rmat using GPU",
+        __FILE__, __LINE__));
+    
+
+    return retval;
+>>>>>>> Stashed changes
 }
 
 /**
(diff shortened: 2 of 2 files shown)


aschaffer:
Potentially, if that offers a minimal disruption compromise for both of us. Can you post some use-case scenario / example? How would we configure that parameter (`grmat`)? How would we use it to "activate" grmat? (I suspect you might already have an example, in which case can you just point me to it?). Thanks.

neoblizz:
> Potentially, if that offers a minimal disruption compromise for both of us. Can you post some use-case scenario / example? How would we configure that parameter (`grmat`)? How would we use it to "activate" grmat? (I suspect you might already have an example, in which case can you just point me to it?). Thanks.

As simple as: `parameters.Set<bool>("grmat", true);`, all the other parameters can be set similarly(`rmat-a/b/c/d`, `graph-nodes`, ...)

neoblizz:
@aschaffer any other feedback before I merge this and release the new version?

aschaffer:
Yes, I posted some questions (perhaps under the wrong PR): https://github.com/gunrock/gunrock/pull/595#issuecomment-533297220
Specifically, `rmat.cuh` removed `Coo_nv`, which we depend on. What did you replace that class with?

neoblizz:
> Yes, I posted some questions (perhaps under the wrong PR): [#595 (comment)](https://github.com/gunrock/gunrock/pull/595#issuecomment-533297220)
> Specifically, `rmat.cuh` removed `Coo_nv`, which we depend on. What did you replace that class with?

> @neoblizz, as a result of replacing `grmat.cuh` with `rmat.cuh` (`grmat.cuh` is still there, but cannot be included because it includes headers that don't exist anymore, like `memset_kernel.cuh`) a lot of functionality that we used to generate graphs has vanished.
> 
> Examples:
> `struct Coo_nv`;
> `BuildRmatGraph_coo_nv(...)`
> 
> What should we replace those with? It looks like our implementation of `gdf_grmat_gen(...)`, the graph generator that uses Gunrock code and residing in `cugraph/cpp/src/utilities/grmat.cu`, must be completely replaced. But with what?
> 
> BTW, here's a list of files that seem to be dead code, since they also include now defunct `memset_kernel.cuh`:
> 
> ```
> $ find ./ -name "*.c*" -exec grep -nH "memset_kernel" '{}' \;
> ./gunrock/app/cc/cc_problem.cuh:18:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/app/dup/dup_partitioner.cuh:24:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/app/cp/cp_partitioner.cuh:23:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/app/sp/sp_partitioner.cuh:23:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/app/brp/brp_partitioner.cuh:23:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/app/rp/rp_partitioner.cuh:24:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/priority_queue/near_far_pile.cuh:19:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/util/kernel_runtime_stats.cuh:17://#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/util/global_barrier.cuh:18:#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/util/array_utils.cuh:25://#include <gunrock/util/memset_kernel.cuh>
> ./gunrock/graphio/grmat.cuh:17:#include <gunrock/util/memset_kernel.cuh>
> (base) aschaffer$ find ./ -name "memset_kernel.cuh"
> (base) aschaffer$
> ```

`Coo_nv`, is being used as just a simple Coordinate graph, you can now use our fully featured coordinate class (that does all sorts of fun stuff, but also has the simple functionality that you need https://github.com/gunrock/gunrock/blob/master/gunrock/graph/coo.cuh). At the core level, it is just storing edge_pairs, edge_values and node_values. So, say if you wanted to generate a grmat graph, just call the load function.

https://github.com/gunrock/gunrock/blob/fc229b445ef67dc007791d544de8d0e03fe47d8e/gunrock/graphio/graphio.cuh#L205

```
parameters.Set<bool>("grmat", true);
GUARD_CU(rmat::Load(parameters, graph, graph_prefix));
```

or if you wanna use a higher level interface: 

```
parameters.Set<bool>("rmat", true); // will use rmat generator
parameters.Set<bool>("grmat", true); // will use GPU for rmat generator
gunrock::graphio::LoadGraph(parameters, graph); // pass the parameters and GraphT type graph
```

To be more specific, the code here:

https://github.com/rapidsai/cugraph/blob/branch-0.10/cpp/src/utilities/grmat.cu#L181-L186

can be changed to look more like this actual example, where CSR graph type is being used, but you can obviously easily switch it to use COO instead:

https://github.com/gunrock/gunrock/blob/master/gunrock/app/sssp/sssp_app.cu#L248-L257

neoblizz:
Addressing other issues:

- grmat file will be deleted, we won't need it with the latest implementation. So, I suggest using our GraphT support and rmat generator.
- I can merge stuff in pretty soon (like by tonight if everything you require is set)

aschaffer:
@neoblizz, I'd say let's merge your changes, create a release, and in the meantime I'll figure out how to change `grmat` on our side to use your new API. Thanks.

neoblizz:
> @neoblizz, I'd say let's merge your changes, create a release, and in the meantime I'll figure out how to change `grmat` on our side to use your new API. Thanks.

That sounds good, let me do that. Also, just to reiterate, the old struct was not at all supported properly, it was a fix to get some stuff working. Now we do a much better job with graph type. The links I posted above should help with the transition (sssp is an example).

neoblizz:
@aschaffer btw, all this code will now be removed as well (all handled in rmat here: https://github.com/gunrock/gunrock/blob/61b55ec1483f1830c6396382560c9f99518dc432/gunrock/graphio/rmat.cuh#L495-L507)

https://github.com/rapidsai/cugraph/blob/branch-0.10/cpp/src/utilities/grmat.cu#L226-L246

aschaffer:
@neoblizz, I have a few Gunrock issues / comments / questions that have just emerged:
1. `subgraphs returns just the _number_ of matches, correct?
2. the `sm(...)` pointers are assumed to be *device* pointers, right?
3. `subgraphs` output: what is its *size*? shouldn't _size_ be returned, too?
4. `row_offsets` is of type *SizeT*. Shouldn't have been of type *VertexT*?
5. The non-template `sm(...)` is still not properly exposed, apparently (unless I'm missing something). Where is its prototype (header declaration; just to confirm we're on the same page: the template `sm(...)` cannot be the declaration of the non-template `sm(...)`;
6. The *template* `sm(...)` function, _at least_, should be inside a gunrock namespace (probably under gunrock::app::sm); to avoid name clashes (sm is too short of a name that can invite name clashes in the future);

aschaffer:
Also, any attempt to instantiate the template `sm(...)` (with anything) triggers compiler errors in gunrock:

2/home/aschaffer/Development/Projects/Graph/forked/fea_ext_gunrock_sm/cugraph/cpp/../thirdparty/gunrock/gunrock/app/s\
   m/sm_app.cuh(228): error: no instance of function template "gunrock::app::TestGraph<_VertexT, _SizeT, _ValueT, _FLAG\
   , _cudaHostRegisterFlag>::FromCsr [with _VertexT=int32_t, _SizeT=int32_t, _ValueT=unsigned long, _FLAG=272U, _cudaHo\
   stRegisterFlag=0U]" matches the argument list
 23            argument types are: (gunrock::graph::Csr<int32_t, int32_t, unsigned long, 272U, 0U, true>, __nv_bool, __\
   nv_bool)
 24            object type is: GraphT
 25          detected during:
 26            instantiation of "double sm(SizeT, SizeT, const SizeT *, const VertexT *, const GValueT *, int, VertexT \
   *) [with VertexT=int32_t, SizeT=int32_t, GValueT=unsigned long]"

jowens:
@Laurawly can you please answer @aschaffer's questions, fix whatever's necessary to make it work, and document the answers in the source code?

aschaffer:
Another important issue that I've just realized. On our side we assumed that subgraph matching means: "the subgraph isomorphism problem is a computational task in which two 
graphs G and H are given as input, and one must determine whether G 
contains a subgraph that is isomorphic to H." (Wikipedia). 

But the `sm(...)` API only takes one graph as input. There should be two of them as input. Otherwise, what is it meant by "subgraph matching" here? Or, what am I missing? Thank you.

aschaffer:
`sm(...)` eventually calls `gunrock_sm(...)` by passing the search graph (the one you search into) and then _loading_ (?) via some parameter logic a query subgraph (the one you search for). How do we pass the query subgraph?

Perhaps `gunrock_sm(...)` should be the one exposed as API, not `sm(...)`, since the latter takes only one graph argument (?) while the former takes 2 graphs?

neoblizz:
> `sm(...)` eventually calls `gunrock_sm(...)` by passing the search graph (the one you search into) and then _loading_ (?) via some parameter logic a query subgraph (the one you search for). How do we pass the query subgraph?
> 
> Perhaps `gunrock_sm(...)` should be the one exposed as API, not `sm(...)`, since the latter takes only one graph argument (?) while the former takes 2 graphs?

I think the `sm(...)` interface should also expose the query graph, that should solve this issue. @Laurawly will be able to answer/address rest of the issues.

Laurawly:
> Another important issue that I've just realized. On our side we assumed that subgraph matching means: "the subgraph isomorphism problem is a computational task in which two
> graphs G and H are given as input, and one must determine whether G
> contains a subgraph that is isomorphic to H." (Wikipedia).
> 
> But the `sm(...)` API only takes one graph as input. There should be two of them as input. Otherwise, what is it meant by "subgraph matching" here? Or, what am I missing? Thank you.

Here's an example of passing query graph: https://github.com/gunrock/gunrock/blob/dev/examples/sm/run.sh#L10 
Pattern graph is query graph.

aschaffer:
> 
> 
> > Another important issue that I've just realized. On our side we assumed that subgraph matching means: "the subgraph isomorphism problem is a computational task in which two
> > graphs G and H are given as input, and one must determine whether G
> > contains a subgraph that is isomorphic to H." (Wikipedia).
> > But the `sm(...)` API only takes one graph as input. There should be two of them as input. Otherwise, what is it meant by "subgraph matching" here? Or, what am I missing? Thank you.
> 
> Here's an example of passing query graph: https://github.com/gunrock/gunrock/blob/dev/examples/sm/run.sh#L10
> Pattern graph is query graph.

That's using it via a shell script. That's _not_ how we intend to use it. We use it programmatically from within a C++ API that further gets exposed to Python. And the C++ API must be as close to "speed of light" as possible. We need a C++ API that can take the source and the query graphs with minimal IO time spent.

Laurawly:
> @neoblizz, I have a few Gunrock issues / comments / questions that have just emerged:
> 
> 1. `subgraphs` returns just the _number_ of matches, correct?
> 2. the `sm(...)` pointers are assumed to be _device_ pointers, right?
> 3. `row_offsets` is of type _SizeT_. Shouldn't have been of type _VertexT_?
> 4. The non-template `sm(...)` is still not properly exposed, apparently (unless I'm missing something). Where is its prototype (header declaration; just to confirm we're on the same page: the template `sm(...)` cannot be the declaration of the non-template `sm(...)`;
> 5. The _template_ `sm(...)` function, _at least_, should be inside a gunrock namespace (probably under gunrock::app::sm); to avoid name clashes (sm is too short of a name that can invite name clashes in the future);

1. Yes, 'subgraphs' returns just the number of matches
2. I'm not sure which pointers you are referring to, if it's the double sm(...) function, then they are host pointers. 
3. VertexT are used for vertex related values such as vertex ids. For offsets sm and other apps use SizeT
4. and 5. sm(...) is calling gunrock_sm which is under gunrock::app::sm namespace. You can suggest a longer name for this app.

Laurawly:
> > > Another important issue that I've just realized. On our side we assumed that subgraph matching means: "the subgraph isomorphism problem is a computational task in which two
> > > graphs G and H are given as input, and one must determine whether G
> > > contains a subgraph that is isomorphic to H." (Wikipedia).
> > > But the `sm(...)` API only takes one graph as input. There should be two of them as input. Otherwise, what is it meant by "subgraph matching" here? Or, what am I missing? Thank you.
> > 
> > 
> > Here's an example of passing query graph: https://github.com/gunrock/gunrock/blob/dev/examples/sm/run.sh#L10
> > Pattern graph is query graph.
> 
> That's using it via a shell script. That's _not_ how we intend to use it. We use it programmatically from within a C++ API that further gets exposed to Python. And the C++ API must be as close to "speed of light" as possible. We need a C++ API that can take the source and the query graphs with minimal IO time spent.

Currently gunrock load query graph via graphio which can be found here: https://github.com/gunrock/gunrock/blob/dev/gunrock/app/sm/sm_app.cuh#L230.  The same way as reading the large graph data. What's the way you are suggesting?

aschaffer:
I am suggesting exposing `gunrock_sm()` (or similar) as API; i.e., a function that takes 2 graphs G (source) and H (query) and returns the number of matches of subgraph H within G. Which seems to be what `gunrock_sm()` is doing. We cannot have IO other than the dataframes dictated by the Python API. Or, if we need such supplementary IO it must be very carefully trimmed, such as to avoid performance penalties.

aschaffer:
Also, there are some header dependencies that trigger linker errors. Specifically, if we include 'sm_app.cuh` header we get linker (!) errors. That's probably because that segment of Gunrock's chain of header dependency references some non-template function(s) that are defined in implementation files (hence, we need to link against). That's probably a sign that a better header separation of templates vs. non-templates could be done in Gunrock, so that we don't need to link. We'd prefer not to link against Gunrock, but rely on a "header-only" implementation. Linking will add too many complications for us at the moment.

jowens:
@Laurawly Can you address @aschaffer's question on which function should be exposed that takes two graphs as input and a count as output?

Laurawly:
@aschaffer Change of exposing gunlock_sm will require change of gunrock Interface since it's currently the same for all apps to not exposing gunrock_xx(...). @jowens I suggest discuss with @sgpyc before we make the decision.

jowens:
@neoblizz might chime in, but I don't see that changing what we expose for SM necessarily means that we have to change what we expose for all other calls. We need to expose the calls that people want to use, and calls that take two graph inputs and output {a list of matches, a count of matches" seem like the right calls to expose.

aschaffer:
@Laurawly, how about exposing a _new_ SM function, for the sake of this conversation call it `sm_correct(...)`, that just takes 2 graphs, as discussed above, and calls `gunrock_sm(...)` inside?

So, the data flow that comes from Python bindings that we have in place is passing 2 graphs. We cannot afford to run shell scripts from Python and much less C++ APIs for obvious performance reasons. So, then  with current `sm(...)` implementation, how can I pass the 2nd (query) graph? (assuming I convert the query graph from our internal format to the Gunrock format; or, how can I just pass the CSR representation of the query graph to `sm(...)`).

Laurawly:
@aschaffer That sounds good to me. I can expose another function. While I can still use the original gunrock interface.

jowens:
Great. @Laurawly please (eventually) expose both a function that returns a count and one that returns the matches.

neoblizz:
> @aschaffer Change of exposing gunlock_sm will require change of gunrock Interface since it's currently the same for all apps to not exposing gunrock_xx(...). @jowens I suggest discuss with @sgpyc before we make the decision.

To clarify, gunrock's interface for c/c++ isn't meant to be set in stone, because for different algorithms we report different things and have different inputs. So, feel free to change that for SM to take in two graphs instead of adding a redundant function wrapper.

neoblizz:
@aschaffer I was looking through some linking stuff, maybe the reason why it isn't working for you is because it needs to be added in the `CMakeLists.txt` here? https://github.com/gunrock/gunrock/blob/master/gunrock/CMakeLists.txt#L30-L44

aschaffer:
@neoblizz, adding `util/str_to_T.cu` fixed the issue. Thank you.

PR #619, opened 2019-10-02
Added wrapper function to pass two graphs to sm

Laurawly:
@aschaffer Could you take a look to see if the added function means what you raised in the issue [here](https://github.com/gunrock/gunrock/pull/606)? @neoblizz Could you check if `typedef` could be used in the way exposed in `nv_sm`. I'll remove the comments once it meets the needs. Thanks!

### DIFF
--- a/gunrock/app/sm/sm_app.cuh
+++ b/gunrock/app/sm/sm_app.cuh
@@ -110,7 +110,7 @@ cudaError_t RunTests(
     }
 
     UseParameters_test(parameters);
-    parameters.Set("num_subgraphs", num_subgraphs);
+    parameters.Set("num-subgraphs", num_subgraphs);
 
     // compute running statistics
     info.ComputeTraversalStats(enactor, (VertexT*)NULL);
--- a/gunrock/app/sm/sm_app.cuh
+++ b/gunrock/app/sm/sm_app.cuh
@@ -205,7 +205,7 @@ double sm(
           VertexT     *subgraphs)
 {
     typedef typename gunrock::app::TestGraph<VertexT, SizeT, GValueT,
-        gunrock::graph::HAS_EDGE_VALUES | gunrock::graph::HAS_CSR>
+        gunrock::graph::HAS_CSR>
         GraphT;
     typedef typename GraphT::CsrT CsrT;
 
@@ -252,14 +252,14 @@ template <
     typename GValueT = unsigned long>
 double nv_sm(
     gunrock::app::TestGraph<VertexT, SizeT, GValueT,
-    gunrock::graph::HAS_EDGE_VALUES | gunrock::graph::HAS_CSR> &query_graph,
+    gunrock::graph::HAS_CSR> &query_graph,
     gunrock::app::TestGraph<VertexT, SizeT, GValueT,
-    gunrock::graph::HAS_EDGE_VALUES | gunrock::graph::HAS_CSR> &data_graph,
+    gunrock::graph::HAS_CSR> &data_graph,
     const int          num_runs,
     VertexT           *subgraphs)
 {
     typedef typename gunrock::app::TestGraph<VertexT, SizeT, GValueT,
-        gunrock::graph::HAS_EDGE_VALUES | gunrock::graph::HAS_CSR>
+        gunrock::graph::HAS_CSR>
         GraphT;
     typedef typename GraphT::CsrT CsrT;
 
--- a/gunrock/app/sm/sm_app.cuh
+++ b/gunrock/app/sm/sm_app.cuh
@@ -71,6 +71,7 @@ cudaError_t RunTests(
     cpu_timer.Stop();
     parameters.Set("preprocess-time", cpu_timer.ElapsedMillis());
 
+    int num_subgraphs = 0;
     // perform SM
     for (int run_num = 0; run_num < num_runs; ++run_num)
     {
@@ -94,7 +95,7 @@ cudaError_t RunTests(
             GUARD_CU(problem.Extract(h_subgraphs));
             SizeT num_errors = app::sm::Validate_Results(
                 parameters, data_graph, query_graph, 
-                h_subgraphs, ref_subgraphs, false);
+                h_subgraphs, ref_subgraphs, &num_subgraphs, false);
         }
     }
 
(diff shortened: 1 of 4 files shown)


aschaffer:
> 
> 
> @aschaffer Could you take a look to see if the added function means what you raised in the issue [here](https://github.com/gunrock/gunrock/pull/606)? @neoblizz Could you check if `typedef` could be used in the way exposed in `nv_sm`. I'll remove the comments once it meets the needs. Thanks!

It looks good. I like the idea of imposing the `HAS_CSR` requirement on `TestGraph` via a boolean non-type template parameter, but I'm a little unclear about the logic of or-ing that with `HAS_EDGE_VALUES`. Does that mean that `TestGraph` is expected to: 
(a) have either a CSR representation or a COO one; 
or, 
(b) have either a CSR representation _or_ it must have edge values?

If it's (b) then what if the caller wants to match a subgraph only based on topology information (neighborhood, adjacency, etc.) without having values for edges? Would the algorithm not work in this case?

neoblizz:
> > @aschaffer Could you take a look to see if the added function means what you raised in the issue [here](https://github.com/gunrock/gunrock/pull/606)? @neoblizz Could you check if `typedef` could be used in the way exposed in `nv_sm`. I'll remove the comments once it meets the needs. Thanks!
> 
> @Laurawly, it looks good. I like the idea of imposing the `HAS_CSR` requirement on `TestGraph` via a boolean non-type template parameter, but I'm a little unclear about the logic of or-ing that with `HAS_EDGE_VALUES`. Does that mean that `TestGraph` is expected to:
> (a) have either a CSR representation or a COO one;
> or,
> (b) have either a CSR representation _or_ it must have edge values?
> 
> If it's (b) then what if the caller wants to match a subgraph only based on topology information (neighborhood, adjacency, etc.) without having values for edges? Would the algorithm not work in this case?

It's a bitwise OR, so it means that the graph must be in CSR representation and must have edge values. (@Laurawly can comment on whether the algorithm can work without the edge values).

When checking for these flags, we do a bitwise AND: `flag & HAS_CSR` for example.

aschaffer:
> It's a bitwise OR, so it means that the graph must be in CSR representation **and** must have edge values. (@Laurawly can comment on whether the algorithm can work without the edge values).
> 
> When checking for these flags, we do a bitwise AND: `flag & HAS_CSR` for example.

Bitwise OR should mean "graph must be in CSR representation **OR** must have edge values." (your statement above says "and"). So, you're saying it should be "AND"?

However, @neoblizz, my question was, essentially, why even impose the "HAS_EDGE_VALES" constraint? That looks like over-constraining to me. I suspect there's a good reason for that. I would just like to understand it. 

But then, we also need to figure out if there's a simple way to make the algorithm work even if the graph DOES NOT have EDGE_VALUES. If there isn't, no worries. But we need to confirm if that is, indeed the case, because the users will ask that, sooner or later.

neoblizz:
> Bitwise OR should mean "graph must be in CSR representation **OR** must have edge values." (your statement above says "and"). So, you're saying it should be "AND"?

No, the OR here is enabling the restrictions on the graph not checking them. Let's say 0001 and 0010 are two different flags, when you bitwise OR them you enable both for that graph: 0001 | 0010 = 0011. Now, let's say I would like to check if the graph has 0010 in it, I would just do a bitwise and : 0011 & 0010.
 
> However, @neoblizz, my question was, essentially, why even impose the "HAS_EDGE_VALES" constraint? That looks like over-constraining to me. I suspect there's a good reason for that. I would just like to understand it.
> 
> But then, we also need to figure out if there's a simple way to make the algorithm work even if the graph DOES NOT have EDGE_VALUES. If there isn't, no worries. But we need to confirm if that is, indeed the case, because the users will ask that, sooner or later.

I understand what you are suggesting here, @Laurawly will be able to help with this.

Laurawly:
> > It's a bitwise OR, so it means that the graph must be in CSR representation **and** must have edge values. (@Laurawly can comment on whether the algorithm can work without the edge values).
> > When checking for these flags, we do a bitwise AND: `flag & HAS_CSR` for example.
> 
> Bitwise OR should mean "graph must be in CSR representation **OR** must have edge values." (your statement above says "and"). So, you're saying it should be "AND"?
> 
> However, @neoblizz, my question was, essentially, why even impose the "HAS_EDGE_VALES" constraint? That looks like over-constraining to me. I suspect there's a good reason for that. I would just like to understand it.
> 
> But then, we also need to figure out if there's a simple way to make the algorithm work even if the graph DOES NOT have EDGE_VALUES. If there isn't, no worries. But we need to confirm if that is, indeed the case, because the users will ask that, sooner or later.

@aschaffer Current sm implementation doesn't require graph to have edge values (edge weights). But if users input edge weights as a matching criteria, we can also accept that. I'm discussing with @neoblizz whether `gunrock::graph::HAS_EDGE_VALUES | gunrock::graph::HAS_CSR` is the correct way doing it.

Laurawly:
@neoblizz I added json ouput support referencing your graph coloring app. Plz lemme know how it looks.

gunrock/app/sm/sm_app.cuh
```diff
@@ -108,6 +109,9 @@ cudaError_t RunTests(
             h_subgraphs, ref_subgraphs, false);
     }
 
+    UseParameters_test(parameters);
+    parameters.Set("num_subgraphs", num_subgraphs);
```
neoblizz: You need to add this in UseParameters first before setting it: https://github.com/gunrock/gunrock/blob/9fe1ef2418ee3a215b2ac439a36c4f32f766c55b/gunrock/app/color/color_app.cu#L42-L45

Also, can you please use `num-subgraphs` and not underscore? All json variables use minus/dash instead

neoblizz: Also, can you please test it with --json to see if the num-subgraphs actually show up?

Laurawly:
@neoblizz I updated the code. I tested the code with `--json` and found the output json has `num-subgraphs = 0`, while I did `printf` just before parameter set, it output the correct value. I'm not sure what's the reason.

neoblizz:
> @neoblizz I updated the code. I tested the code with `--json` and found the output json has `num-subgraphs = 0`, while I did `printf` just before parameter set, it output the correct value. I'm not sure what's the reason.

You must be using the boost version of json, which is now removed in master (and dev I think). If you force it to use rapidjson, that should work.

Laurawly:
@neoblizz Are we good in merging this PR?

neoblizz:
Yes.

PR #621, opened 2019-10-08
Gunrock Shared Build:: HITS/SM...

jdwapman:
- [X] Implement an API and implementation for the HITS C interface

- [x] Solve "multiple definition" compile errors when attempting to use more than one app enabled in CMakeLists.txt (HITS and SM, specifically)

- [ ] Test HITS C API correctness [FUTURE]


jdwapman:
@neoblizz This should pass all tests and add the multiple-app fix to Gunrock. I'll add a test of my HITS interface in a future commit after I resolve a segfault issue.

PR #637, opened 2019-10-30
[CI] Google test added for subgraph matching

Laurawly:
@neoblizz Please review.


Laurawly:
> What happened here with pr? @Laurawly @neoblizz

Two apps compiled together here gives duplicated definition error as follows:
```
nvlink error   : Multiple definition of '_ZN7gunrock5oprtr12intersection5ilog2Ej' in '/home/laurawly/workplace/gunrock/build/gunrock/CMakeFiles/gunrock.dir/app/sm/./gunrock_generated_sm_app.cu.o', first defined in '/home/laurawly/workplace/gunrock/build/gunrock/CMakeFiles/gunrock.dir/app/pr/./gunrock_generated_pr_app.cu.o'
nvlink fatal   : Internal error: duplicate relocations at same address
```
@neoblizz mentioned this is a known issue and @jdwapman has an active PR which is trying to fix it. So he suggested that we test the subgraph matching first app and make sure everything works.

jowens:
Got it. Thank you.

jdwapman:
I ran into some segfault issues implementing my HITS interface but I can make a new pull request with the fixes, and add my test in a subsequent pull request once I resolve them.

Laurawly:
> I ran into some segfault issues implementing my HITS interface but I can make a new pull request with the fixes, and add my test in a subsequent pull request once I resolve them.

Thanks, that'll be great.

jdwapman:
Pending pull request updated with the fixes. @neoblizz good to merge?

neoblizz:
> Pending pull request updated with the fixes. @neoblizz good to merge?

Thanks, merged the other one. After #638, we can merge this. @Laurawly can you make sure you uncomment all the PR stuff? The changes that are now going live don't require us to only run one app, all will work together.

Laurawly:
Thanks @jdwapman. @neoblizz I'll update my pr now.

neoblizz:
@Laurawly I am gonna merge #638, even though its broken, because I think the fixes in this PR will fix the sm stuff broken in 638.

neoblizz:
@Laurawly change this:
https://github.com/gunrock/gunrock/blob/f57471241765ba6326868ba87ced8c42f5be21ba/gunrock/app/sm/sm_app.cu#L28-L31
to 
```
GUARD_CU(parameters.template Use<unsigned int>(
              "num-subgraphs",
              util::REQUIRED_ARGUMENT | util::SINGLE_VALUE | util::INTERNAL_PARAMETER,
              0, "number of matched subgraphs", __FILE__, __LINE__));
```

neoblizz:
@Laurawly you could also try and keep your old version instead of using `ParametersT` as a template parameter. Because this should be under `sm::UseParameters`, so you should not get any duplicated definition of that function.

gunrock/app/sm/sm_app.cu
```diff
@@ -17,15 +17,14 @@ namespace gunrock {
...
-cudaError_t UseParameters(ParametersT &parameters) {
+cudaError_t UseParameters(util::Parameters &parameters) {
   cudaError_t retval = cudaSuccess;
   GUARD_CU(UseParameters_app(parameters));
   GUARD_CU(UseParameters_problem(parameters));
   GUARD_CU(UseParameters_enactor(parameters));
   GUARD_CU(UseParameters_test(parameters));
 
-  GUARD_CU(parameters.Use<unsigned int>(
+  GUARD_CU(parameters.template Use<unsigned int>(
```
neoblizz: Sorry for the misunderstanding, you don't need `.template` if you switch it back to the old version. But if it compiles, you can leave this in there.

Laurawly: The CI passed but I can change it back if you think it's necessary.

Laurawly:
@aschaffer I updated the interface in this PR for people to call subgraph matching by sending in data graph and query graph by sending their row offsets and column indices. For example, you can reference a toy example here: https://github.com/Laurawly/gunrock/blob/4fe863c0e79e63df681f88666514cb4c2b08adc4/unittests/test_lib_sm.h.

Issue #616, opened 2019-09-26
HITS in gunrock.h


jdwapman:
Need to expose HITS in gunrock.h

> Also, another issue: HITS. It doesn't seem exposed in `gunrock.h` (maybe I'm wrong). Is that work in progress?

> @jdwapman can you please work on exposing this in `gunrock.h`?

_Originally posted by @neoblizz in https://github.com/gunrock/gunrock/pull/595#issuecomment-529689366_

PR #648, opened 2019-11-15
Implemented a working HITS shared lib test

jdwapman:
Adds a functional C interface to gunrock.h and implements a shared library test for the HITS algorithm.

This pull request is in response to #616

### DIFF
--- a/examples/shared_libs/shared_lib_hits.c
+++ b/examples/shared_libs/shared_lib_hits.c
@@ -1,0 +1,30 @@
+/**
+ * @brief PageRank test for shared library advanced interface
+ * @file shared_lib_pr.c
+ */
+
+#include <stdio.h>
+#include <gunrock/gunrock.h>
+
+int main(int argc, char *argv[]) {
+  ////////////////////////////////////////////////////////////////////////////
+
+  int num_nodes = 7, num_edges = 26;
+  int row_offsets[8] = {0, 3, 6, 11, 15, 19, 23, 26};
+  int col_indices[26] = {1, 2, 3, 0, 2, 4, 0, 1, 3, 4, 5, 0, 2,
+                         5, 6, 1, 2, 5, 6, 2, 3, 4, 6, 3, 4, 5};
+
+  // HITS
+  float *hub_ranks = (float *)malloc(sizeof(float) * num_nodes);
+  float *auth_ranks = (float *)malloc(sizeof(float) * num_nodes);
+  int num_iter = 10;
+  double elapsed_hits = hits(num_nodes, num_edges, row_offsets, col_indices, num_iter, hub_ranks, auth_ranks);
+
+  for (int node = 0; node < num_nodes; ++node)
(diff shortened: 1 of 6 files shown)


examples/shared_libs/shared_lib_hits.c
```diff
@@ -0,0 +1,30 @@
+/**
+ * @brief PageRank test for shared library advanced interface
+ * @file shared_lib_pr.c
```
neoblizz: Update comments please to `hits`.

gunrock/app/hits/hits_app.cuh
```diff
@@ -205,21 +201,21 @@ double gunrock_hits(
 }
```
Laurawly: What if we have number of runs larger than 1 here? Will `gunrock_hits` capture all the runtime?

gunrock/app/hits/hits_app.cuh
```diff
@@ -249,11 +246,13 @@ double hits(
     data_graph.CsrT::row_offsets   .SetPointer((SizeT *)row_offsets, num_nodes + 1, gunrock::util::HOST);
     data_graph.CsrT::column_indices.SetPointer((VertexT *)col_indices, num_edges, gunrock::util::HOST);
```
Laurawly: This can also be replaced by target if target is defined before the above two lines.

jdwapman: I'll update this, thanks

gunrock/app/hits/hits_app.cuh
jdwapman: No other apps have a parameter to specify a number of runs greater than 1. num_iter can be used to specify how many HITS iterations to perform, since this is an iterative algorithm.

gunrock/app/hits/hits_app.cuh
```diff
@@ -205,21 +201,21 @@ double gunrock_hits(
...
  * @param[in]  row_offsets CSR-formatted graph input row offsets
  * @param[in]  col_indices CSR-formatted graph input column indices
- * @param[in]  edge_values CSR-formatted graph input edge weights
- * @param[in]  num_runs    Number of runs to perform SM
- * @param[out] subgraphs   Return number of subgraphs
- * \return     double      Return accumulated elapsed times for all runs
+ * @param[in]  num_iter    Number of iterations to perform HITS
+ * @param[out] hub_ranks   Vertex hub scores
+ * @param[out] auth ranks  Vertex authority scores
+ * \return     double      Return accumulated elapsed times for all iterations
```
Laurawly: Here you mentioned returning accumulated elapsed times for all iterations but you only run one.

gunrock/app/hits/hits_app.cuh
Laurawly: Other apps have the option of changing the number of runs to multiple. And in gunrock_xx they are accumulating the total time ran by `num_runs`.

neoblizz: That's not particularly important, because you could always have a loop outside the run and do the cpu_timer accumulation yourself at the c call level.

PR #652, opened 2019-11-20
Shared Nearest Neighbor (improved/refactored)

agalup:
Separate KNN from SNN to two applications.
Changed representation of input points from graph with edges to the set of points with n-dim labels.
Provided new function to parsing points to one n-dim array in labels.cuh. Some other functions added as well because of one n-dim array on input. Another function in labels.cuh are not changed.
Fixed algorithm of SNN: previous version was not classical version according to [1][2] - Core and CPU_Reference changed according this.
Changed Validation: the same cluster can got different id on CPU and GPU - now this case is considered.
Added helpsers to KNN. Compute_dist function to compute distance between two points. Right now it is euclidean metric.
Added helpers to SNN. New __device__ function to computing SNNsimilarity of two points. New Union Find data structure to join clusters in CPU_Reference.
Added new generator of tests, which generates "stars" - 5-points stars which can overlap. 
Clean up debug.


examples/snn/snn outputs
```diff
@@ -0,0 +1,247 @@
+agnes@bowser:~/knn_gunrock/tests/knn$ make && ./bin/test_knn_10.0_x86_64 market ../../dataset/small/chesapeake.mtx --k=4 --eps=1 --min-pts=3
```
neoblizz: Remove the output log files from the PR.

gunrock/app/knn/knn_app.cu
```diff
@@ -39,24 +42,29 @@ cudaError_t UseParameters(util::Parameters &parameters) {
...
+      util::REQUIRED_ARGUMENT | util::REQUIRED_PARAMETER, 
+      "", "List of points of dim-dimensional space", __FILE__, __LINE__));
 
   GUARD_CU(parameters.Use<int>(
-      "x",
-      util::REQUIRED_ARGUMENT | util::MULTI_VALUE | util::OPTIONAL_PARAMETER, 0,
-      "Index of reference point.", __FILE__, __LINE__));
+      "dim",
+      util::REQUIRED_ARGUMENT | util::MULTI_VALUE | util::OPTIONAL_PARAMETER,
+      0, "Dimensional of space", __FILE__, __LINE__));
```
neoblizz: typo: "Dimensions"

neoblizz: Also, why is the default 0?

agalup: I will change to 2

neoblizz:
> It is a copy of function which was in knn_helpers, is it needed here?

What do you mean? It was missing from the commits so I just added the file and cleaned it up.

neoblizz:
> This is not correct.
> Values of h_knns[v_k+i], ref_knns[v_k + i] can be different and solution can be still correct. Two points can be different and their distance from given "v" can be the same. For example: let say we have 3 points: 1, 2, 3, that distance(1,2)=distance(1,3). If k=1 (we are looking for only 1 nearest neighbor) then either 2 or 3 could be knn of 1.

In validate, I was avoiding computing the distance again and comparing that instead. Ideally, if you use CPU reference code, I should be able to say "these are the K nearest neighbors for point X on CPU and on GPU, and they are equal, therefore, algorithm is correct", no?

agalup:
> > It is a copy of function which was in knn_helpers, is it needed here?
> 
> What do you mean? It was missing from the commits so I just added the file and cleaned it up.

I mean that exactly the same function is in knn_helpers so we can just link to knn_helpers. Also this function is not called from test_knn (knn_test has link to knn_helpers). Also metric could be change if needed so it supposed to be in one place only and not duplicate, right?

agalup:
> > This is not correct.
> > Values of h_knns[v_k+i], ref_knns[v_k + i] can be different and solution can be still correct. Two points can be different and their distance from given "v" can be the same. For example: let say we have 3 points: 1, 2, 3, that distance(1,2)=distance(1,3). If k=1 (we are looking for only 1 nearest neighbor) then either 2 or 3 could be knn of 1.
> 
> In validate, I was avoiding computing the distance again and comparing that instead. Ideally, if you use CPU reference code, I should be able to say "these are the K nearest neighbors for point X on CPU and on GPU, and they are equal, therefore, algorithm is correct", no?

h_knns and ref_knns is keeping index of point, not value of distance between two points. This example which I put above is showing that you could have different results for two algorithms and both of them are correct. If k=1 - one nearest neighbor is needed, then any of points: 2 or 3 is correct because both are on the same distance from point 1.

neoblizz:
> knns is keeping index of point, not v

I see what you mean, so, as long as the distance is the same (minimum) we can confirm validation. The previous code SegFaulted with the distance check, that is another reason why I got rid of it.

neoblizz:
@agalup I am merging this, let's open a new PR for SNN fixes.

PR #653, opened 2019-11-20
Add boost cpu reference for subgraph matching

Laurawly:
@neoblizz @porumbes Could you help review?

### DIFF
--- a/gunrock/app/sm/sm_app.cuh
+++ b/gunrock/app/sm/sm_app.cuh
@@ -233,12 +233,17 @@ double sm_template(
     data_graph.CsrT::Allocate(num_nodes, num_edges, gunrock::util::HOST);
     data_graph.CsrT::row_offsets   .SetPointer((SizeT *)row_offsets, num_nodes + 1, gunrock::util::HOST);
     data_graph.CsrT::column_indices.SetPointer((VertexT *)col_indices, num_edges, gunrock::util::HOST);
-    data_graph.FromCsr(data_graph.csr(), true, quiet);
+
+    gunrock::util::Location target = gunrock::util::HOST;
+    data_graph.FromCsr(data_graph.csr(), target, 0, quiet, true);
     gunrock::graphio::LoadGraph(parameters, data_graph);
-    data_graph.CsrT::Allocate(num_query_nodes, num_query_edges, gunrock::util::HOST);
-    data_graph.CsrT::query_row_offsets   .SetPointer((SizeT *)query_row_offsets, num_query_nodes + 1, gunrock::util::HOST);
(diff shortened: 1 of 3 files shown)


gunrock/app/sm/sm_test.cuh
```diff
@@ -87,6 +95,77 @@ void DisplaySolution(T *array, SizeT length) {
...
+               typename CorrespondenceMap2To1>
+      bool operator()(CorrespondenceMap1To2 f, CorrespondenceMap2To1) {
+          // Count number of matches from isomorphism map
+          (*count_ptr)++;
+          return true;
+      }
+  private:
+      const Graph1& graph1_;
+      const Graph2& graph2_;
+      size_t *count_ptr{nullptr};
```
porumbes: I'm not sure I understand why the `count_ptr` variable exists? Why not increment `count_` directly?

Laurawly: 'Cause boost copy the callback structure multiple times, it's impossible to accumulate count. I tried to set count_ as a static member, but boost doesn't allow me to do that either.

Laurawly: I have a script for you if you wanna try it out: https://gist.github.com/Laurawly/7da2bfc2e0610705c7d3db3c57229c11

porumbes: Ah, I see. You should be able to get rid of the `count_ptr` stuff if you use `std::ref` like so `  vf2_subgraph_iso(graph1, graph2, std::ref(callback));`

Laurawly: yeah I can do ref as well.

gunrock/app/sm/sm_app.cuh
```diff
@@ -229,16 +229,21 @@ double sm(
     bool quiet = parameters.Get<bool>("quiet");
     GraphT data_graph;
     GraphT query_graph;
+
+    gunrock::util::Location target = gunrock::util::HOST;
     // Assign pointers into gunrock graph format
     data_graph.CsrT::Allocate(num_nodes, num_edges, gunrock::util::HOST);
```
porumbes: Let's make this allocation on `target` too.

gunrock/app/sm/sm_test.cuh
```diff
@@ -156,7 +154,7 @@ double CPU_Reference(util::Parameters &parameters, GraphT &data_graph,
 
       // Print out all mappings
```
porumbes: Please fix this comment. I don't think `callback` is  printing anything out here.

Issue #593, opened 2019-09-08
Replace moderngpu with CUB


jowens:
@neoblizz helpfully found all mgpu instances. CUB is likely better supported going forward.

```
./gunrock/app/enactor_types.cuh:238:  mgpu::ContextPtr context;
./gunrock/app/enactor_types.cuh:269:      context = mgpu::CreateCudaDeviceAttachStream(gpu_idx, stream);
./gunrock/oprtr/LB_CULL_advance/kernel.cuh:1096:    mgpu::SortedSearch<mgpu::MgpuBoundsLower>(
./gunrock/oprtr/advance/advance_base.cuh:181:  mgpu::Scan<mgpu::MgpuScanTypeInc>(
./gunrock/oprtr/advance/advance_base.cuh:184:      (SizeT)0, mgpu::plus<SizeT>(),
./gunrock/oprtr/advance/kernel.cuh:103:(Value)0, mgpu::plus<typename KernelPolicy::Value>(), context); break;
./gunrock/oprtr/advance/kernel.cuh:108:(Value)1, mgpu::multiplies<typename KernelPolicy::Value>(), context); break;
./gunrock/oprtr/advance/kernel.cuh:113:(Value)INT_MIN, mgpu::maximum<typename KernelPolicy::Value>(), context); break;
./gunrock/oprtr/advance/kernel.cuh:118:(Value)INT_MAX, mgpu::minimum<typename KernelPolicy::Value>(), context); break;
./gunrock/oprtr/advance/kernel.cuh:124:(Value)0, mgpu::plus<typename KernelPolicy::Value>(), context); break;
./gunrock/oprtr/advance/kernel.cuh:224: * @tparam Op Operation for gather reduce. mgpu::plus<int> by default.
./gunrock/oprtr/filter/kernel.cuh:208:    Scan<mgpu::MgpuScanTypeExc>(
./gunrock/oprtr/filter/kernel.cuh:210:        mgpu::plus<SizeT>(), (SizeT *)NULL, (SizeT *)NULL,
./gunrock/oprtr/oprtr_parameters.cuh:48:  mgpu::ContextPtr context;
./gunrock/oprtr/LB_advance/kernel.cuh:564:    mgpu::SortedSearch<mgpu::MgpuBoundsLower>(
./gunrock/oprtr/LB_advance/kernel.cuh:600:  (Value)0, mgpu::plus<typename KernelPolicy::Value>(), context); break;
[...]

PR #658, opened 2019-11-27
Code clean and mgpu::scan replace with cub::scan issue #593

yzhwang:
1) Replace mgpu::Scan with cub scan.
2) Remove all compile-time warnings, leave only the ones from rapidjson.


gunrock/app/louvain/louvain_test.cuh
```diff
@@ -1107,7 +1101,7 @@ typename GraphT::SizeT Validate_Results(
 
   for (VertexT v = 0; v < graph.nodes; v++) {
     auto c = ref_communities[v];
-    if (c < 0 || c >= graph.nodes) {
+    if (util::lessThanZero(0) || c >= graph.nodes) {
```
neoblizz: Should be `util::lessThanZero(c)`

gunrock/oprtr/advance/advance_base.cuh
```diff
@@ -143,49 +113,34 @@ template <OprtrFlag FLAG, typename GraphT, typename FrontierInT,
...
-      (SizeT *)NULL,
-      parameters.frontier->output_offsets.GetPointer(util::DEVICE),
-      parameters.context[0]);
+
+  util::cubInclusiveSum(parameters.frontier->cub_temp_space,
+                        /*d_in=*/parameters.frontier->output_offsets,
+                        /*d_out=*/parameters.frontier->output_offsets,
+                        /*num_items=*/parameters.frontier->queue_length);
+  GUARD_CU2(
+          cudaMemcpy(parameters.frontier->output_length.GetPointer(util::DEVICE),
```
neoblizz: There should be an Array1D utility for this copy, I believe you should just be able to do `.Move` after setting the pointer, and the move will be from `util::DEVICE` to `util::DEVICE`. We'd like to avoid using direct cuda APIs in the code without the wrapper.

neoblizz:
> Remove all compile-time warnings, leave only the ones from rapidjson.
Would you happen to know how we can address the warnings from rapidjson? I tried using rapidjson with clang and it didn't seem to have these warnings (idk what it was). I also had an active issue for this: https://github.com/Tencent/rapidjson/issues/1545

gunrock/app/louvain/louvain_test.cuh
yzhwang: sorry for the typo. Will fix.

gunrock/oprtr/advance/advance_base.cuh
yzhwang: Yeah I would really like to avoid. But Move() only works on same Array1D object? I checked there is no such Move() function. Or did I miss it? We can of course add one later to provide Move() between Array1D objects. I will add a TODO.

yzhwang:
On rapidjson warning: Yeah saw your issue. Unfortunately have no clue how to avoid in gcc.

gunrock/oprtr/advance/advance_base.cuh
neoblizz: No, it could be used across any pointer. At least we do that for all Extract functions in problem by setting the pointer of an Array1D to a new location, and then initiating a move on it. Let me look it up to make sure.

neoblizz: It may be missing `source == DEVICE && target == DEVICE`, I didn't get a chance to properly go over it: https://github.com/gunrock/gunrock/blob/master/gunrock/util/array_utils.cuh#L771-L884

neoblizz: Or, like you said, we could always write it and have it a `TODO`.

yzhwang: Thanks for the link! In this context I actually want to do a d2d memcpy between two Array1D objects. I think a simple function like this: https://github.com/moderngpu/moderngpu/blob/master/src/moderngpu/memory.hxx#L38 would do the trick. Our Move() is missing a pointer to other Array1D object in its interface.
Let me put a TODO and I can add this in later commit.
Another option is to directly send in the device pointer to total length in scan. That would save an explicit memcpy outside the kernel but alas, we need to change cub implementation.

yzhwang:
Accidentally added my unfinished k-core commits to this PR... Let me fix this first...

yzhwang:
Will just close this PR and do a new one. Sorry for my rusty git skill...

PR #659, opened 2019-11-27
Shared Nearest Neighbor (implementation, correction and optimizations)

neoblizz:


neoblizz:
@agalup https://github.com/gunrock/gunrock/blob/fb5ea4f0939cf20e06d7efa1d1d915015b4066ce/examples/knn/test_knn.cu#L172 this is also a problem if our points are only `double` type. We should just force it to be `F64` for the time being.

CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ include(${CMAKE_SOURCE_DIR}/cmake/FindOpenMP.cmake)
 if (METIS_LIBRARY)
   include(${CMAKE_SOURCE_DIR}/cmake/FindMetis.cmake)
 endif()
+include(${CMAKE_SOURCE_DIR}/cmake/FindFAISS.cmake)
```
neoblizz: Have you tested if FAISS was not found, does the cmake still build properly?

examples/knn/test_knn.cu
```diff
@@ -169,7 +167,7 @@ int main(int argc, char** argv) {
 
   app::Switch_Types<app::VERTEXT_U32B | app::VERTEXT_U64B |
                            app::SIZET_U32B | app::SIZET_U64B |
-                           app::VALUET_F32B | app::UNDIRECTED>(
+                           app::VALUET_F64B | app::UNDIRECTED>(
```
neoblizz: Is it possible to enable both F32B and F64B now? Because we may wanna run `float` and `double` for future. As FAISS' KNN is only `float` (32-bit support).

examples/snn/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
...
+#  Gunrock: Sub-Project Shared Nearest Neighbor
+# ------------------------------------------------------------------------
+project(snn)
+message("-- Project Added: ${PROJECT_NAME}")
+include(${CMAKE_SOURCE_DIR}/cmake/SetSubProject.cmake)
+
+target_link_libraries(${PROJECT_NAME} ${CUDA_cudadevrt_LIBRARY})
+
+if (FAISS_INCLUDE_DIR)
+    include_directories(${PROJECT_NAME} ${FAISS_INCLUDE_DIR})
```
neoblizz: You already do this in `SetSubProject.cmake`. Should that one be removed and just keep the one in `snn/CMakeLists.txt`? `SetSubProject.cmake` is included in here.

examples/snn/Makefile
```diff
@@ -1,4 +1,4 @@
-# ----------------------------------------------------------------
+#1----------------------------------------------------------------
```
neoblizz: remove `1`

examples/snn/Makefile
```diff
@@ -13,6 +13,22 @@ include ../BaseMakefile.mk
 ALGO = snn
 EXEC = bin/test_$(ALGO)_$(NVCC_VERSION)_$(ARCH_SUFFIX)
 
+FAISS_INC =
+FAISS_LINK =
+
+ifdef FAISS_INCLUDE_DIR
```
neoblizz: Where is `FAISS_INCLUDE_DIR`? CMakeLists =/= BaseMakefile.mk, they don't see each other's variables. I think here you have to do something like what we do for `use_metis=1`. Check BaseMakefile.mk for example.

examples/snn/run.sh
```diff
@@ -17,35 +17,69 @@ do
...
+PATH[6]="/home/agnes/clustering_dataset/Berton/dim3"
+PATH[7]="/home/agnes/clustering_dataset/Berton/dim3"
+PATH[8]="/home/agnes/clustering_dataset/Bower/dim3"
+PATH[9]="/home/agnes/clustering_dataset/FOF"
+PATH[10]="/home/agnes/clustering_dataset/FOF"
+PATH[11]="/home/agnes/clustering_dataset/FOF"
+PATH[12]="/home/agnes/clustering_dataset/FOF"
+PATH[13]="/home/agnes/clustering_dataset/FOF"
+PATH[14]="/home/agnes/clustering_dataset/FOF"
+PATH[15]="/home/agnes/clustering_dataset/FOF"
```
neoblizz: Let's update these paths to `/data/clustering_dataset/` and I will add all the other datasets to that main path.

examples/snn/run_tests_json.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
```
neoblizz: We can remove this file.

examples/snn/test_snn.cu
```diff
@@ -12,17 +12,31 @@
...
+    #include <faiss/gpu/GpuResources.h>
+    #include <faiss/gpu/StandardGpuResources.h>
+    #include <faiss/utils/Heap.h>
+    #include <faiss/gpu/utils/Limits.cuh>
+    #include <faiss/gpu/utils/Select.cuh>
+#endif
 
 // JSON includes
-#include <gunrock/util/info_rapidjson.cuh>
+//#include <gunrock/util/info_rapidjson.cuh>
```
neoblizz: Remove.

gunrock/app/knn/knn_helpers.cuh
```diff
@@ -24,9 +24,53 @@ namespace knn {
 
 template <typename ValueT>
 __device__ __host__ ValueT _sqrt(const ValueT &a) {
-    return (ValueT) sqrtf((float)a);
+    return sqrt(a);
 }
```
neoblizz: We may have to support both float and double for this:

```c++
template <typename ValueT>
__device__ __host__ ValueT _sqrt(const float &a) {
    return sqrtf((float)a);
}

template <typename ValueT>
__device__ __host__ ValueT _sqrt(const double &a) {
    return sqrt((double)a);
}
```

gunrock/oprtr/1D_oprtr/for_all.cuh
```diff
@@ -50,6 +50,36 @@ __global__ void ForAll_Kernel(ArrayT array, ApplyLambda apply, SizeT length) {
...
+  const SizeT STRIDE = (SizeT)blockDim.x * gridDim.x;
+  SizeT i = (SizeT)blockDim.x * blockIdx.x + threadIdx.x;
+  while (i < length) {
+/*    if (blockDim.x * blockIdx.x + threadIdx.x == 0){
+        printf("%d points done\n", i);
+    }*/
+    apply(array + 0, i);
+    i += STRIDE;
+  }
+}
```
neoblizz: Why does debug kernel exist here? If this is solely used for debugging, it should be a test. I don't think it should be added to the main API.

CMakeLists.txt
agalup: Yes, I tested it. Also Jenkins tested it. There is no FAISS on daisy.

neoblizz:
@agalup we should prepare to merge this.

crozhon:
Hey is there anything preventing this from being merged? Anything I can help out with? I'm happy to review too if that's needed.

examples/snn/Makefile
agalup: FAISS_INCLUDE_DIR is environment variable. It has to be defined from console before compilation

examples/snn/run.sh
agalup: I am going to remove this file. Those paths does not exists anymore

gunrock/app/knn/knn_helpers.cuh
agalup: This does not work

agalup:
I think it is done now. Please merge PR if changes are right.

gunrock/app/knn/knn_helpers.cuh
neoblizz: I am sorry, why does this not work?

examples/snn/Makefile
neoblizz: Can you add this comment to the Makefile? Just so we know for future?

neoblizz:
> I think it is done now. Please merge PR if changes are right.

Just two minor comments! This looks great.

gunrock/app/knn/knn_helpers.cuh
agalup: How shall I deal with the case when I have:
```
ValueT x = fun();
  _sqrt(x);
```
? The above function does not work on it (compiler shows up error). Why I cannot leave how is right now?

examples/snn/Makefile
agalup: It is already there. While compilation it says: "FAISS not found. In order to use FAISS, please define FAISS_INCLUDE_DIR"

gunrock/app/knn/knn_helpers.cuh
neoblizz: I thought `sqrt` was for `double` only, so you'd feed to cast it to a double then if you want to use it for any `ValueT`.

agalup: If it is for double only then I do not need the first function you posted above, right? Do you want me to do only for doubles and then cast each value to double before using `_sqrt`?

PR #660, opened 2019-11-28
Code Clean, Replace mgpu::Scan() with cub scan.

yzhwang:


yzhwang:
Adding some note:
Using cub scan will cause illegal memory access when queue length is oversize and need to expand queue length. I need to investigate more. For now please hold the merge. Thanks.

neoblizz:
> Adding some note:
> Using cub scan will cause illegal memory access when queue length is oversize and need to expand queue length. I need to investigate more. For now please hold the merge. Thanks.

Just a note on this, I have a version of gunrock working with mgpu 2.0, which is much easier to upkeep with the latest cuda arch. Worth considering that in the short term and making sure mgpu is up-to-date, rather than using CUB. I know the original issue is to remove mgpu all together, but just a remark on that.

yzhwang:
> > Adding some note:
> > Using cub scan will cause illegal memory access when queue length is oversize and need to expand queue length. I need to investigate more. For now please hold the merge. Thanks.
> 
> Just a note on this, I have a version of gunrock working with mgpu 2.0, which is much easier to upkeep with the latest cuda arch. Worth considering that in the short term and making sure mgpu is up-to-date, rather than using CUB. I know the original issue is to remove mgpu all together, but just a remark on that.

Agree on that. Might even consider upgrade to that version? I'm (slowly) working on adding new primitives to mgpu 2.0 too. Did a multi-scan sum-array-table implementation (originally in cudpp, very useful for image processing apps) and a sum by index/ segmented {intersection,...} on the way. I mean I'm definitely much more familiar with mgpu than cub. OK back to this issue, might consider closing this and just put the code clean part in? Should I do that?

neoblizz:
> > > Adding some note:
> > > Using cub scan will cause illegal memory access when queue length is oversize and need to expand queue length. I need to investigate more. For now please hold the merge. Thanks.
> > 
> > 
> > Just a note on this, I have a version of gunrock working with mgpu 2.0, which is much easier to upkeep with the latest cuda arch. Worth considering that in the short term and making sure mgpu is up-to-date, rather than using CUB. I know the original issue is to remove mgpu all together, but just a remark on that.
> 
> Agree on that. Might even consider upgrade to that version? I'm (slowly) working on adding new primitives to mgpu 2.0 too. Did a multi-scan sum-array-table implementation (originally in cudpp, very useful for image processing apps) and a sum by index/ segmented {intersection,...} on the way. I mean I'm definitely much more familiar with mgpu than cub. OK back to this issue, might consider closing this and just put the code clean part in? Should I do that?

Yeah, let's do that. And consider moving towards mgpu 2.0 instead -- the calls are just an easy replacement, because it takes all the features from mgpu 1.0 (even for the scan, you don't need to do the copy like you do for cub).

neoblizz:
And work on updating it to support SM_60/65/70 (Volta).

PR #664, opened 2019-12-06
Market IO improvement

yzhwang:
Minor perf improvement by switching to std::getline().

An alternative solution would be to read into a giant buffer than use omp to parse in parallel.

Also a possible enhancement is to make arbitrary parser in callback functions.


neoblizz:
> An alternative solution would be to read into a giant buffer than use omp to parse in parallel.

This may play a huge role in loading large graphs with limited sever times.

> Also a possible enhancement is to make arbitrary parser in callback functions.
Could you elaborate on that? Worth creating an enhancement issue for these.

PR #665, opened 2019-12-10
Compile for detected compute capabilities

crozhon:
Gunrock currently compiles for practically all possible compute
capabilites, which requires extremely long build times. It also causes
several warnings to be generated when these various combos of options
go the compiler. This change greatly reduces the amount of code we
normally compile in development, and reduces the complexity of build
configurations.

The file AutoDetectCudaArch.cmake was written by Ahmed Mahmoud.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,17 +256,17 @@ option(GUNROCK_GENCODE_SM52
 if (CUDA_VERSION VERSION_EQUAL "8.0" OR CUDA_VERSION VERSION_GREATER "8.0")
   option(GUNROCK_GENCODE_SM60
     "ON to generate code for Compute Capability 6.0 devices (e.g. Tesla P100)"
-    ON)
+    OFF)
   option(GUNROCK_GENCODE_SM61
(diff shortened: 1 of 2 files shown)


crozhon:
Looking at the Jenkins results, this takes a typical build from 17mins to 7 mins, so it should hopefully improve productivity for folks.

jowens:
@crozhon did you write this? This is cool. Even if you stole it from elsewhere, this is still cool.

crozhon:
It’s modified from Ahmed’s gpu_mesh project: https://github.com/owensgroup/gpu_mesh/blob/master/cmake/AutoDetectCudaArch.cmake

Thanks!

neoblizz:
@crozhon so, I am assuming our `cuda_properties.cuh` was ok? Did we also try it on an RTX card?

crozhon:
@neoblizz Just tried it out on an RTX card, and I still got the warnings. By building only the required compute capabilities, I only hid them on daisy. I'm working further on the warnings now (some of the math in the kernel.cuh / kernel_policy.cuh files is off now), but this PR which brings the build time down can still come in.

PR #673, opened 2020-01-07
Remove hard-coded and ambiguous CUDA_ARCH

crozhon:
Many operators currently do occupancy calculations using the variable
CUDA_ARCH. This value is hard-coded in multiple places to be 300.
This change disambiguates multiple overlapping uses of the variable,
and ensures that it uses the currently compiling compute capability.
This allows occupancy calculations to use the correct numbers and
eliminates some build warnings when compiling with RTX cards.

Some of the KernelPolicy structs defined by operators use a template
variable _CUDA_ARCH (note the leading underscore). These variables
are hard-coded to 300 by their template typedefs. Other KernelPolicy
structs simply use CUDA_ARCH, which will pick up the static
definition in oprtr_base.h

All uses of CUDA_ARCH will now point to a global value GR_CUDA_ARCH
defined in cuda_properties.cuh rather than have this ambiguous
behavior. It also sets this value based on the currently compiling
compute capability. When compiling host device code, it will use the
value of 300, as the actual value should be irrelevant during these
passes.

### DIFF
--- a/gunrock/util/cuda_properties.cuh
+++ b/gunrock/util/cuda_properties.cuh
@@ -14,21 +14,26 @@
 
 #pragma once
 
-namespace gunrock {
-namespace util {
-
 /*****************************************************************
  * Macros for guiding compilation paths
  *****************************************************************/
 
 /**
  * CUDA architecture of the current compilation path
+ *
+ * When compiling nvcc sets the __CUDA_ARCH__ environment variable
+ * to the current compute capability being compiled for.
+ * We set GR_CUDA_ARCH to this value if it's available and 300
+ * otherwise (in host-side code)
  */
-#ifndef __CUDA_ARCH__
-//#define __GR_CUDA_ARCH__ 0                      // Host path
+#ifdef __CUDA_ARCH__ // Device-side compilation
+#define GR_CUDA_ARCH __CUDA_ARCH__
 #else
-#define __GR_CUDA_ARCH__ __CUDA_ARCH__  // Device path
+#define GR_CUDA_ARCH 300 // Host-side compilation
 #endif
+
+namespace gunrock {
+namespace util {
 
 /*****************************************************************
  * Device properties by SM architectural version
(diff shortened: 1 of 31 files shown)


crozhon:
This should allow a performance gain since our occupancy calculations always assumed compute capability = 300. There's further work to be done here detecting the occupancy numbers at run-time with new CUDA APIs (https://devblogs.nvidia.com/cuda-pro-tip-occupancy-api-simplifies-launch-configuration/) instead of all this template magic at compile time.

jowens:
Eager to see if this helps performance! Looks like a lot of low-level excavating on your part, kudos.

crozhon:
I've added a comment in cuda_properties.cuh where GR_CUDA_ARCH is defined to better explain it.

KP_CUDA_ARCH was from when I was testing which CUDA_ARCHs were which (so just for debugging). Jenkins and I didn't catch it because it appears that edge_map_backwards/... isn't used anywhere.

gunrock/util/cuda_properties.cuh
```diff
@@ -14,22 +14,27 @@
...
+ * We set GR_CUDA_ARCH to this value if it's available and 300
+ * otherwise (in host-side code)
  */
-#ifndef __CUDA_ARCH__
-//#define __GR_CUDA_ARCH__ 0                      // Host path
+#ifdef __CUDA_ARCH__ // Device-side compilation
+#define GR_CUDA_ARCH __CUDA_ARCH__
 #else
-#define __GR_CUDA_ARCH__ __CUDA_ARCH__  // Device path
+#define GR_CUDA_ARCH 300 // Host-side compilation
```
neoblizz: Why `300` and is it necessary to have this at all for host-side compilation?

crozhon: It's only necessary to get the templates in KernelPolicy structs to compile during the host-side pass. Any number will work, but it just has to exist so that math inside the templates can work. I chose 300 because that's what was used host-side before. 

I agree that it's a hack, so I'd like to either:
- Move the occupancy calculations out of templates and use the occupancy APIs that CUDA provides
- Not compile the structs host side (and so don't do the math at all) with some #ifdefing 

I thought for now it'll get the job done though.

neoblizz: Ok, that's a fair explanation. Let's make an issue about this or document this in an existing issue so we remember to get back to it when we switch to use CUDA's occupancy calculator.

PR #677, opened 2020-01-11
Upgrading Gunrock's dependency to moderngpu 2.0

neoblizz:
Preparing for the refactor, for which I use moderngpu 2 instead of 1. Hoping to get some a nice performance improvements with the following changes and the fixes issued in PR #673.

## TODO
- [x] Update moderngpu's [launchbox ](https://github.com/moderngpu/moderngpu/wiki/Launching-kernels#launch-box) to support `SM > 53`
- [x] Bug somewhere in a shfl_sync primitive of moderngpu causing a hang in some apps
- [x] Silence the output of `mgpu::context_t()`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,7 +5,7 @@
 [submodule "externals/moderngpu"]
 	path = externals/moderngpu
 	url = https://github.com/NVlabs/moderngpu.git
-	branch = branch_1.1
+	branch = master
 [submodule "docs"]
 	path = docs
 	url = https://github.com/gunrock/docs.git
--- a/cmake/FindModernGPU.cmake
+++ b/cmake/FindModernGPU.cmake
@@ -2,11 +2,11 @@
 #  Gunrock: Find external moderngpu directories
 # ------------------------------------------------------------------------
 SET(mgpu_INCLUDE_DIRS
-  ${CMAKE_SOURCE_DIR}/externals/moderngpu/include
+  ${CMAKE_SOURCE_DIR}/externals/moderngpu/src
   CACHE PATH
   "Directory to the Modern GPU include files")
 
 SET(mgpu_SOURCE_DIRS
-  ${CMAKE_SOURCE_DIR}/externals/moderngpu/src
+  ${CMAKE_SOURCE_DIR}/externals/moderngpu/src/moderngpu
   CACHE PATH
   "Directory to the Modern GPU source files")
--- a/cmake/SetSubProject.cmake
+++ b/cmake/SetSubProject.cmake
@@ -18,8 +18,7 @@ else()
 endif()
 
 set (mgpu_SOURCE_FILES
-  ${mgpu_SOURCE_DIRS}/mgpucontext.cu
-  ${mgpu_SOURCE_DIRS}/mgpuutil.cpp)
+  ${mgpu_SOURCE_DIRS}/context.hxx)
 # end /* moderngpu include directories */
 
 # begin /* CUB include directories */
--- a/examples/BaseMakefile.mk
+++ b/examples/BaseMakefile.mk
@@ -54,7 +54,7 @@ SM_TARGETS = $(GEN_SM70)
 #-------------------------------------------------------------------------------
 
 CUDA_INC = -I"$(shell dirname $(NVCC))/../include"
-MGPU_INC = -I"../../externals/moderngpu/include"
+MGPU_INC = -I"../../externals/moderngpu/src"
 CUB_INC = -I"../../externals/cub"
 
 BOOST_INC =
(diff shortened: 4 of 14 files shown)


crozhon:
For upgrading moderngpu to use SM 75, I've updated the launch_box.hxx and launch_params.hxx files, but I see that several of moderngpu's included algorithms pick values for a certain arch. Like this in kernel_scan

https://github.com/moderngpu/moderngpu/blob/4cd4f486c7ed25d82e5c85cc026d0ca676a6746f/src/moderngpu/kernel_scan.hxx#L18-L22

@neoblizz Do these need to be updated as well? So that we can get the most performance on a card with a higher arch? Or is our performance just coming from updating moderngpu from v1 --> v2? I am not fully understanding the launch boxes though, so that could be the source of my confusion.

jowens:
Making sure @seanbaxter is in this conversation if he wants to be (please see previous comment, Sean).

neoblizz:
> For upgrading moderngpu to use SM 75, I've updated the launch_box.hxx and launch_params.hxx files, but I see that several of moderngpu's included algorithms pick values for a certain arch. Like this in kernel_scan
> 
> https://github.com/moderngpu/moderngpu/blob/4cd4f486c7ed25d82e5c85cc026d0ca676a6746f/src/moderngpu/kernel_scan.hxx#L18-L22
> 
> @neoblizz Do these need to be updated as well? So that we can get the most performance on a card with a higher arch? Or is our performance just coming from updating moderngpu from v1 --> v2? I am not fully understanding the launch boxes though, so that could be the source of my confusion.

An expert's opinion is definitely important here (@seanbaxter), but some of this also trial and error (@yzhwang maybe you can advice on this as well). Per the launch box documentation I linked above:

> The launch box mechanism in moderngpu 1.0 has been refactored and is more powerful and flexible than ever. One of the philosophies of the improved moderngpu is to push tuning out of the library and to the user. The library cannot anticipate every use, every type that its functions are specialized over, every operator used, or every length and distribution of data. It's not feasible to choose constants for each library function that provides the best performance for every variant.

I am not sure if there are defaults that we can call "good-enough" for now for different architectures, before we go around adding our own for each and every algorithm in the library.

`arch_xx_cta` parameter structs derive the `launch_cta_t` primitive, which defines four parameters:
```cpp
template<int nt_, int vt_ = 1, int vt0_ = vt_, int occ_= 0>
struct launch_cta_t {
  enum { nt = nt_, vt = vt_, vt0 = vt0_, occ = occ_ };
};
```
`nt` controls the size of the CTA (thread block) in threads. `vt` is the number of values per thread (grain size). `vt0` is the number of unconditional loads of input made in cooperative parallel kernels (usually set to `vt` for regularly parallel functions and `vt - 1` for load-balancing search functions). `occ` is the occupancy argument of the `__launch_bounds__` kernel decorator. It specifies the minimum CTAs launched concurrently on each SM. 0 is the default, and allows the register allocator to optimize away spills by allocating many registers to hold live state. Setting a specific value for this increases occupancy by limiting register usage, but potentially causes spills to local memory. `arch_xx_cta` structs support all four arguments, although `nt` is the only argument required.

seanbaxter:
Use trial and error. There is no model for choosing tunings, especially since nvidia doesn't provide performance information on their SMs. The only thing to keep in mind is you'll need to choose an odd values-per-thread constant if you want to avoid bank conflicts during transposes. I don't think migrating from mgpu 1 to mgpu 2 should require more tuning.. It's more likely that the simpler design of mgpu 2 pushes the launch box closer to the user, forcing you to look at it and consider occupancy.

Don't trust the defaults I put in. Good tunings are application-specific and vary by data type, number of concurrent streams, and so on.

neoblizz:
> Use trial and error. There is no model for choosing tunings, especially since nvidia doesn't provide performance information on their SMs. The only thing to keep in mind is you'll need to choose an odd values-per-thread constant if you want to avoid bank conflicts during transposes. I don't think migrating from mgpu 1 to mgpu 2 should require more tuning.. It's more likely that the simpler design of mgpu 2 pushes the launch box closer to the user, forcing you to look at it and consider occupancy.
> 
> Don't trust the defaults I put in. Good tunings are application-specific and vary by data type, number of concurrent streams, and so on.

Thank you Sean, super helpful! Looking forward to updating moderngpu 2.0 to support the latest stuff. Excited to see some performance numbers!

crozhon:
@seanbaxter Thank you so much Sean! I especially appreciate the tip about the odd values-per-thread constant. I understand now that the launch boxes have been pushed to the user, and I appreciate the interface you've given us. 

Yuxin has also provided the fix for the issue with sync, which I'm taking a closer look at now.

yzhwang:
The initial comment did cause some hang. I fixed it in a PR (did this a while a go but forgot to open the PR): https://github.com/moderngpu/moderngpu/pull/36

neoblizz:
Thanks to Sean, the bug-fix commit is merged to moderngpu. @crozhon can you update the moderngpu submodule to the latest commit (master branch of moderngpu)?

crozhon:
Updated moderngpu to latest. 

I am specifying our launch_boxes before our uses of mgpu. Just putting values there for now that seem sensible and seem to work. This is just about ready. 

I also have a PR just about ready for updating moderngpu's launch_box to consider SM > 75.

crozhon:
I have a PR for those small changes to the definitions: 
https://github.com/moderngpu/moderngpu/pull/37

crozhon:
I've updated the PR to include the launch_boxes for Scan and SortedSearch, which appear to be the only mgpu primitives we use. I've also silenced the output of standard_context_t.

Issue #629, opened 2019-10-22
PageRank problem


OsipovMax:
**Summary:**
1. If I run a PageRank test with an RMAT graph type and also specify a graph-scale parameter value greater than 5, I get "CUDA error 77: an illegal memory access was encountered".
2. If I specify the graph-edgefactor parameter, I also get "CUDA error 77: an illegal memory access was encountered".
But if I don't specify the graph-edgefactor parameter and the graph-scale parameter value is less than 6, then everything works fine in this case.

P.S. On other tests, there is no such problem.

**Reproducibility:**
Always

**To Reproduce:**
сlone gunrock
mkdir build && cd build
cmake ..
make pr
cd bin
./pr --graph-scale 20 --graph-type rmat --graph-edgefactor 32

**Expected Behavior:**
Generate pr result.

**Actual Results:**
CUDA error 77: an illegal memory access was encountered.

**System Configuration:**
GPU: NVIDIA Tesla P100
Driver Version: 418.40.04
CUDA Version: 10.1

neoblizz:
Related to #615

PR #682, opened 2020-01-21
Fix for pagerank illegal memory access

porumbes:
The fix is in `pr_test.cuh` (was missing `util::DEVICE` in `new_coo` allocation).

Runs like: `./pr --graph-scale 20 --graph-type rmat --graph-edgefactor 32` (from duplicate issue #629) complete as expected.

The additional commits add a utility (and associated gtest) that helped me generate some stack traces during debugging.


gunrock/app/pr/pr_test.cuh
```diff
@@ -94,7 +94,7 @@ cudaError_t Compensate_ZeroDegrees(GraphT &graph, bool quiet = false) {
   CooT new_coo;
   GUARD_CU(new_coo.Allocate(graph_coo.nodes + 1,
                             graph_coo.edges + counter + graph_coo.nodes,
-                            util::HOST));
+                            util::HOST | util::DEVICE));
```
neoblizz: Wow, good catch!

PR #678, opened 2020-01-11
Fix bugs for subgraph matching

Laurawly:
This pr fixes the correction issue for query graph other than triangles. It uses combination method to store intermediate results iteration by iteration.

PR #685, opened 2020-01-21
Update google test for subgraph matching

Laurawly:
Update google test for subgraph matching to make count to be the number of matched subgraphs. Changing according to the merge of PR #678


neoblizz:
@crozhon can you please review this? Merge if everything is good.

PR #691, opened 2020-02-06
Initial JSON writer improvements

crozhon:
Draft pull request for changes to JSON writer. I've already done some testing locally, so I'd just appreciate another pair of eyes on these changes. 

TODO:
- [X] Never write an invalid file  
- [X] Sort keys before printing file 
- [x] Write a file at start containing the command line and other information, but perhaps mark the elapsed time as -1. What's the best way to denote a failed run?

### DIFF
--- a/gunrock/util/info_rapidjson.cuh
+++ b/gunrock/util/info_rapidjson.cuh
@@ -14,12 +14,16 @@
 
 #pragma once
 
+#define RAPIDJSON_HAS_STDSTRING 1
+
 #include <cstdio>
 #include <vector>
 #include <ctime>
+#include <cmath>
 #include <time.h>
 #include <rapidjson/prettywriter.h>
 #include <rapidjson/filewritestream.h>
+#include <rapidjson/document.h>
 #include <gunrock/util/gitsha1.h>
 #include <gunrock/util/sysinfo_rapidjson.h>
 
@@ -56,11 +60,6 @@ cudaError_t UseParameters_info(ParametersT &parameters) {
       "", "Tag to better describe and identify json outputs", __FILE__,
       __LINE__));
 
-  // GUARD_CU(parameters.template Use<uint64_t>(
-  //   "filtered-srcs",
-  //   util::REQUIRED_ARGUMENT | util::MULTI_VALUE | util::INTERNAL_PARAMETER, 0,
-  //   "Array of filtered source vertices", __FILE__, __LINE__));
-
   return retval;
 }
 
@@ -88,7 +87,6 @@ struct Info {
   int64_t edges_queued;
   double nodes_redundance;
   double edges_redundance;
-  // double     load_time;
   double preprocess_time;
   double postprocess_time;
   double write_time;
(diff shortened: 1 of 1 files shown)


neoblizz:
Once you have all the stuff sorted out -- let's try and generate a complete run for one of the algorithms to make sure we are getting proper output. (I recommend PR, since that one had issues)

crozhon:
Found out the PR issue. Some double or float values can be Inf or NaN. RapidJSON disallows these because they're not in the spec. So I decided to write them as strings instead, i.e. `"nodes_redundance": "inf"` 

Let me know if you think other behavior is better. Otherwise everything is great. Below is an example PR run that is interrupted, as well as one that goes through normally. Note the avg-processed-time key. 

[pr_complete.json.txt](https://github.com/gunrock/gunrock/files/4168087/pr_complete.json.txt)
[pr_incomplete.json.txt](https://github.com/gunrock/gunrock/files/4168088/pr_incomplete.json.txt)

jowens:
Concern is that if you write a string for something that ought to be a number, a parser (e.g., a python parser) is going to think that field is a string field, and that's a giant pain if you're trying to, say, plot the data in that field. Also, parsers get confused if the field is sometimes a string and sometimes a number.

jowens:
There is a JSON `null` value (not `"null"`), maybe that is most appropriate? (Check me on that.)

crozhon:
> Concern is that if you write a string for something that ought to be a number, a parser (e.g., a python parser) is going to think that field is a string field, and that's a giant pain if you're trying to, say, plot the data in that field. Also, parsers get confused if the field is sometimes a string and sometimes a number.

I see. I imagine leaving the key out would also break things? Then how about `"nodes_redundance": null`

crozhon:
Haha I see we reached the same conclusion. The Python docs say that would get turned into None when parsed. Does that seem alright from your end?

jowens:
So I would rather the following happen:
1. Some things do not semantically belong in a particular JSON (for instance, `mark-pred` in a PageRank run). In this case, the key shouldn't be in the JSON at all.
2. If something does belong in a JSON but we don't have data for it / we have invalid data, `null` makes sense.
Does that seem like a decent philosophy? I think python will work fine with either missing keys or `null` values for existing keys.

crozhon:
Yes that seems good to me. I was just talking about NaN and Infinity. Those will get turned into null. 

For mark-pred, I am not planning on setting it to null or anything like that. I believe that the PR app as written chooses to add mark-pred to the map in the util::Parameters class, even though it's never actually set or used for anything.

jowens:
OK, good, I like null.

I think I'd rather have something like PR for which `mark-pred` has no meaning _not_ put that in the map at all. But, no biggie either way. Just don't want app writers having to fill in a long list of fields they don't care about.

crozhon:
Yeah that makes sense. I agree and I’m just trying to find the location where PR does put it in the map, so I can remove that line and make sure it only happens for the apps where it has meaning. I think maybe our base_app class puts it in the map for no reason, but I have to hunt it down.

crozhon:
Alright. I've done some more testing and "mark-pred" doesn't show up in PR json's anymore. If you see any other key like that let me know.

PR #709, opened 2020-03-10
Prevent duplication of key "load-time" in output json

achalagarwal:
@crozhon  I still haven't figured out where the second ```setVal()``` for **load-time** is being called, but commenting out this one works!


crozhon:
Could you just delete it instead of commenting it out? We can always find it via git history if needed. Then I'll accept this.

achalagarwal:
Done!

Also, I can't link this pull request to the issue, do I need to be a collaborator to do that? @jowens can you add me?

jowens:
Done.

achalagarwal:
> Done.

Thanks!

But still can't link it to the issue :thinking:

PR #716, opened 2020-04-21
Add target device option in subgraph matching

Laurawly:
Giving users an option of whether to store inputs/outputs in the CPU or GPU. Also output list of subgraphs besides the counting of subgraphs.

### DIFF
--- a/examples/sm/test_sm.cu
+++ b/examples/sm/test_sm.cu
@@ -67,7 +67,7 @@ struct main_struct {
         !quiet);
 
     // counts of matched subgraphs
-    VertexT *ref_subgraph_match = new VertexT[data_graph.nodes];
+    unsigned long *ref_subgraph_match = new unsigned long[data_graph.nodes];
     if (!quick) {
       util::PrintMsg("__________________________", !quiet);
 
--- a/gunrock/app/sm/sm_app.cu
+++ b/gunrock/app/sm/sm_app.cu
@@ -63,8 +63,8 @@ double sm(
     const int                *query_col_indices,
     const int                 num_runs,
     gunrock::util::Location   device,
-          int                *subgraphs,
-          int                *list_subgraphs)
+    unsigned long            *subgraphs,
+    unsigned long            *list_subgraphs)
 {
     return sm_template(num_nodes, num_edges, row_offsets, col_indices,
         num_query_nodes, num_query_edges, query_row_offsets,
--- a/gunrock/app/sm/sm_app.cu
+++ b/gunrock/app/sm/sm_app.cu
@@ -47,9 +47,9 @@ cudaError_t UseParameters(util::Parameters &parameters) {
  * @param[in]  query_row_offsets CSR-formatted graph input query row offsets
  * @param[in]  query_col_indices CSR-formatted graph input query column indices
  * @param[in]  num_runs          Number of runs to perform SM
- * @param[in]  device            input and output target device
  * @param[out] subgraphs         Return number of subgraphs
- * @param[in]
+ * @param[out] list_subgraphs    Return list of subgraphs
+ * @param[in]  device            input and output target device, by default CPU
  * \return     double            Return accumulated elapsed times for all runs
  */
 double sm(
@@ -62,13 +62,13 @@ double sm(
     const int                *query_row_offsets,
     const int                *query_col_indices,
     const int                 num_runs,
-    gunrock::util::Location   device,
     unsigned long            *subgraphs,
-    unsigned long            *list_subgraphs)
+    unsigned long            *list_subgraphs,
+    gunrock::util::Location   device = gunrock::util::HOST)
 {
     return sm_template(num_nodes, num_edges, row_offsets, col_indices,
         num_query_nodes, num_query_edges, query_row_offsets,
-        query_col_indices, num_runs, device, subgraphs, list_subgraphs);
+        query_col_indices, num_runs, subgraphs, list_subgraphs, device);
 }
 
 // Leave this at the end of the file
--- a/gunrock/app/sm/sm_app.cu
+++ b/gunrock/app/sm/sm_app.cu
@@ -47,7 +47,9 @@ cudaError_t UseParameters(util::Parameters &parameters) {
  * @param[in]  query_row_offsets CSR-formatted graph input query row offsets
  * @param[in]  query_col_indices CSR-formatted graph input query column indices
  * @param[in]  num_runs          Number of runs to perform SM
+ * @param[in]  device            input and output target device
  * @param[out] subgraphs         Return number of subgraphs
+ * @param[in]
  * \return     double            Return accumulated elapsed times for all runs
  */
 double sm(
@@ -60,11 +62,12 @@ double sm(
     const int           *query_row_offsets,
     const int           *query_col_indices,
     const int            num_runs,
+       string            device,
           int           *subgraphs)
 {
     return sm_template(num_nodes, num_edges, row_offsets, col_indices,
         num_query_nodes, num_query_edges, query_row_offsets,
-        query_col_indices, num_runs, subgraphs);
+        query_col_indices, num_runs, device, subgraphs);
 }
 
 // Leave this at the end of the file
--- a/gunrock/app/sm/sm_app.cu
+++ b/gunrock/app/sm/sm_app.cu
@@ -49,7 +49,7 @@ cudaError_t UseParameters(util::Parameters &parameters) {
  * @param[in]  num_runs          Number of runs to perform SM
  * @param[out] subgraphs         Return number of subgraphs
  * @param[out] list_subgraphs    Return list of subgraphs
- * @param[in]  device            input and output target device, by default CPU
+ * @param[in]  allocated_on      Input and output target device, by default CPU
  * \return     double            Return accumulated elapsed times for all runs
  */
 double sm(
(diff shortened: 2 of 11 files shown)


jowens:
@Laurawly Can you give an indication of "more testings needed" ... what works, what is expected to work but hasn't been tested, what doesn't work?

gunrock/app/sm/sm_app.cu
```diff
@@ -60,11 +62,12 @@ double sm(
     const int           *query_row_offsets,
     const int           *query_col_indices,
     const int            num_runs,
+       string            device,
```
ChuckHastings, line 65: Is there a reason to use a string here rather than creating an enum that's exposed through the API?

string is an immutable object, copying it down through the call stack will be much more expensive than an enum (which internally is an integer).

gunrock/app/sm/sm_problem.cuh
```diff
@@ -460,9 +475,18 @@ struct Problem : ProblemBase<_GraphT, _FLAG> {
...
+      // returning results will be stored on the CPU
+      if (device == "CPU") {
+        count_subgraphs[0] = combinations.size();
+        list_subgraphs = new VertexT[combinations.size()];
+        std::copy(combinations.begin(), combinations.end(), list_subgraphs);
+      } else { // returning results will be stored on the GPU
+        VertexT *h_subgraphs = new VertexT[1];
+        h_subgraphs[0] = combinations.size();
+        GUARD_CU(cudaMemcpy(count_subgraphs, h_subgraphs, 1, cudaMemcpyHostToDevice));
+        GUARD_CU(cudaMalloc(list_subgraphs, combinations.side()));
```
ChuckHastings, line 487: So this suggests that the internal algorithm is generating the results in host memory and you're going to copy it back out into device memory?  While that will achieve my goal, wouldn't it be more efficient overall if the results were generated directly into GPU memory?

I haven't worked through the logic in the implementation, but at some point there is a CUDA kernel that is iterating over the CSR representation of the graph and generating a sequence of subgraph matches.  If you're storing them in host memory, aren't you paying a transfer cost (which we typically assume is expensive) to bring those subgraphs out of the CUDA kernel into host memory... and now we'll pay that cost again to put it back into device memory.

It would seem that generating the list of subgraph matches directly in device memory would be more efficient.

neoblizz: Yes, you are correct. We should in fact just return the subgraphs that already exist on the device and are accessible in the problem::extract().

gunrock/app/sm/sm_app.cu
neoblizz: @Laurawly we can use the location enum that already exists under `util::Location`, you can name it whatever, but we don't need string here.

Laurawly: Sure, I can change accordingly.

neoblizz:
```
CMakeFiles/shared_lib_sm.dir/shared_lib_sm.c.o: In function `main':
shared_lib_sm.c:(.text+0x157): undefined reference to `sm'
collect2: error: ld returned 1 exit status
```

Do not change from VertexT to ulong, the problem is in the shared lib file that is calling the SM app. Since the interface changed, it needs to be updated now @Laurawly.

Laurawly:
> ```
> CMakeFiles/shared_lib_sm.dir/shared_lib_sm.c.o: In function `main':
> shared_lib_sm.c:(.text+0x157): undefined reference to `sm'
> collect2: error: ld returned 1 exit status
> ```
> 
> Do not change from VertexT to ulong, the problem is in the shared lib file that is calling the SM app. Since the interface changed, it needs to be updated now @Laurawly.
> 
> I am guessing here: https://github.com/gunrock/gunrock/blob/dev/examples/shared_libs/shared_lib_sm.c

Gotcha, I forgot to update that file. Will change accordingly.

Laurawly:
@neoblizz In `gunrock.h`, how can I use gunrock::util::Location for `device` param? Do I need to define a new enum there?

examples/sm/test_sm.cu
```diff
@@ -67,7 +67,7 @@ struct main_struct {
         !quiet);
 
     // counts of matched subgraphs
-    VertexT *ref_subgraph_match = new VertexT[data_graph.nodes];
+    unsigned long *ref_subgraph_match = new unsigned long[data_graph.nodes];
```
neoblizz, line 70: Please change this back to `VertexT`.

gunrock/app/sm/sm_app.cu
```diff
@@ -48,23 +48,27 @@ cudaError_t UseParameters(util::Parameters &parameters) {
...
+    const int                *row_offsets,
+    const int                *col_indices,
+    const int                 num_query_nodes,
+    const int                 num_query_edges,
+    const int                *query_row_offsets,
+    const int                *query_col_indices,
+    const int                 num_runs,
+    unsigned long            *subgraphs,
+    unsigned long            *list_subgraphs,
+    gunrock::util::Location   device = gunrock::util::HOST)
```
neoblizz, line 67: `device = HOST` is confusing. device generally implies GPU.
Can we change this to `target`, or `allocated_on`?

gunrock/app/sm/sm_app.cu
```diff
@@ -48,23 +48,27 @@ cudaError_t UseParameters(util::Parameters &parameters) {
...
 {
     return sm_template(num_nodes, num_edges, row_offsets, col_indices,
         num_query_nodes, num_query_edges, query_row_offsets,
-        query_col_indices, num_runs, subgraphs);
+        query_col_indices, num_runs, subgraphs, list_subgraphs, device);
```
neoblizz, line 71: Same thing here.

gunrock/app/sm/sm_app.cuh
```diff
@@ -44,7 +44,7 @@ cudaError_t UseParameters(util::Parameters &parameters);
  */
 template <typename GraphT, typename VertexT = typename GraphT::VertexT>
 cudaError_t RunTests(util::Parameters &parameters, GraphT &data_graph,
-                     GraphT &query_graph, VertexT *ref_subgraphs,
+                     GraphT &query_graph, unsigned long *ref_subgraphs,
```
neoblizz, line 47: Change back to VertexT.

gunrock/app/sm/sm_app.cuh
```diff
@@ -132,17 +133,21 @@ cudaError_t RunTests(util::Parameters &parameters, GraphT &data_graph,
...
+ * @param[out] list_subgraphs  Return list of matched subgraph combinations
+ * @param[in]  device          Target device where inputs and outputs are stored
+ * \return     double          Return accumulated elapsed times for all runs
  */
 template <typename GraphT, typename ValueT = typename GraphT::ValueT>
 double gunrock_sm(gunrock::util::Parameters &parameters, GraphT &data_graph,
-                  GraphT &query_graph, typename GraphT::VertexT *subgraphs) {
+                  GraphT &query_graph, unsigned long *subgraphs,
+                  unsigned long *list_subgraphs,
+                  gunrock::util::Location device = gunrock::util::HOST) {
```
neoblizz, line 150: Naming convention here as well, maybe change it to `allocated_on`?

gunrock/app/sm/sm_app.cuh
```diff
@@ -195,7 +202,8 @@ double sm_template(const SizeT num_nodes, const SizeT num_edges,
                    const SizeT num_query_nodes, const SizeT num_query_edges,
                    const SizeT *query_row_offsets,
                    const VertexT *query_col_indices, const int num_runs,
-                   VertexT *subgraphs) {
+                   unsigned long *subgraphs, unsigned long *list_subgraphs,
+                   gunrock::util::Location device = gunrock::util::HOST) {
```
neoblizz, line 206: Naming convention here as well, maybe change it to `allocated_on`?

gunrock/app/sm/sm_test.cuh
```diff
@@ -126,7 +126,7 @@ struct vf2_count_callback {
 // TODO: change CPU reference code to count subgraphs instead of triangles
 template <typename GraphT, typename VertexT = typename GraphT::VertexT>
 double CPU_Reference(util::Parameters &parameters, GraphT &data_graph,
-                     GraphT &query_graph, VertexT *subgraphs) {
+                     GraphT &query_graph, unsigned long *subgraphs) {
```
neoblizz, line 129: VertexT

gunrock/app/sm/sm_test.cuh
```diff
@@ -265,33 +265,30 @@ double CPU_Reference(util::Parameters &parameters, GraphT &data_graph,
...
  */
 template <typename GraphT, typename VertexT = typename GraphT::VertexT>
 typename GraphT::SizeT Validate_Results(util::Parameters &parameters,
                                         GraphT &data_graph, GraphT &query_graph,
-                                        VertexT *h_subgraphs,
-                                        VertexT *ref_subgraphs,
-                                        int *num_subgraphs,
+                                        unsigned long *h_subgraphs,
+                                        unsigned long *ref_subgraphs,
+                                        unsigned long *num_subgraphs,
```
neoblizz, line 283: Fix types

examples/sm/test_sm.cu
Laurawly: Because the corresponding variable on the GPU has to be unsigned long type, here I cannot change back to VertexT.

gunrock/app/sm/sm_app.cu
Laurawly: Sure, I can change the name to `allocated_on`

gunrock/app/sm/sm_app.cu
Laurawly: Same type problem

gunrock/app/sm/sm_app.cuh
Laurawly: Same

gunrock/app/sm/sm_app.cuh
Laurawly: Sure

gunrock/app/sm/sm_test.cuh
Laurawly: Same here

gunrock/app/sm/sm_test.cuh
Laurawly: Same

PR #717, opened 2020-04-25
Adding simple advance and iterators

maawad:
Fixed bugs:
* `GraphFlag` problem is fixed
* Bug in insertion edges counters


maawad:
Feedback from @neoblizz and @crozhon would be useful here too. We discussed that we needed to avoid capturing CSR specific arrays like [`weights`](https://github.com/gunrock/gunrock/blob/dynamic-graph/gunrock/app/sssp/sssp_enactor.cuh#L95). One solution is passing the edge/node struct (check the API document) and getting edge/node metadata. I'm currently just passing the [edge value](https://github.com/maawad/gunrock/blob/simple-advance/gunrock/oprtr/simple_advance/kernel.cuh#L87) because its the easiest given what code I have but I'm open to any ideas here.

gunrock/graph/csr_parallel_iterator.cuh
```diff
@@ -0,0 +1,66 @@
...
+ * @tparam VertexT Vertex identifier type.
+ * @tparam SizeT Graph size type.
+ * @tparam ValueT Associated value type.
+ * @tparam GraphFlag graph flag
+ */
+template <typename VertexT, typename SizeT, typename ValueT, GraphFlag FLAG>
+struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_CSR> {
+  using CsrT =
+      graph::Csr<VertexT, SizeT, ValueT, FLAG & graph::HAS_CSR_MASK,
+                 cudaHostRegisterDefault, (FLAG & graph::HAS_CSR) != 0>;
```
porumbes, line 35: I think I'm good with this version, but just curious. Is there any chance that we can simplify by having less files (e.g., this file) if this `struct ParallelIterator` lives inside `struct Csr` in csr.cuh? Would doing so make using this iterator more or less cumbersome?

gunrock/graph/dynamic_graph/dynamic_graph_base.cuh
```diff
@@ -46,14 +51,36 @@ struct DynamicGraphBase {
       typename std::conditional<REQUIRE_EDGES_VALUES, SlabHashGraphMapT,
                                 SlabHashGraphSetT>::type;
 
-  DynamicGraphT dynamicGraph;
+  mutable DynamicGraphT dynamicGraph;
```
porumbes, line 54: What's the `const` function that mutates `dynamicGraph` or why is this `mutable`?

gunrock/graph/dynamic_graph/dynamic_graph_parallel_iterator.cuh
```diff
@@ -0,0 +1,162 @@
...
+  __device__ static void readNode(const uint32_t& key, VertexT& eId,
+                                  ValueT& eVal, const int& laneId) {
+    bool key_lane = laneId < 30;
+    eId = key_lane ? key : EMPTY_KEY;
+    bool is_valid = (key != EMPTY_KEY);
+    eId = (key != EMPTY_KEY) ? eId
+                             : util::PreDefinedValues<VertexT>::InvalidValue;
+    eVal = util::PreDefinedValues<ValueT>::InvalidValue;
+  }
+};
```
porumbes, line 52: Are other users expected to use these or are they for low level internal use only? Should they be part of a private interface?

gunrock/graph/dynamic_graph/dynamic_graph_parallel_iterator.cuh
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * @brief ParallelIterator iterator for dynamic graph.
+ *
+ * @tparam VertexT Vertex identifier type.
+ * @tparam SizeT Graph size type.
+ * @tparam ValueT Associated value type.
+ * @tparam GraphFlag graph flag
+ */
+template <typename VertexT, typename SizeT, typename ValueT, GraphFlag FLAG>
+struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_DYN> {
```
porumbes, line 63: Similar to Csr, does this file need to exist? Does anything get simpler if the iterator lived inside the thing it iterates over?

gunrock/graph/dynamic_graph/kernels/map/insert.cuh
```diff
@@ -38,20 +38,19 @@ __device__ void insertWarpEdges(PairT& thread_edge, ValueT& thread_value,
...
     if (laneId == 0) {
       atomicAdd(&d_edges_per_vertex[cur_src], added_count);
     }
+    if (success) {
+      SizeT bucket_offset = d_buckets_offset[cur_src];
+      atomicAdd(&d_edges_per_bucket[bucket_offset + dst_bucket], 1);
+    }
   }
 }
 template <typename PairT, typename ValueT, typename SizeT, typename ContextT>
```
porumbes, line 56: Looks like the same code in `set/insert.cuh`.

gunrock/graph/dynamic_graph/kernels/set/insert.cuh
```diff
@@ -37,18 +37,18 @@ __device__ void insertWarpEdges(PairT& thread_edge, ContextT*& hashContexts,
...
     if (laneId == 0) {
       atomicAdd(&d_edges_per_vertex[cur_src], added_count);
     }
+    if (success) {
+      SizeT bucket_offset = d_buckets_offset[cur_src];
+      atomicAdd(&d_edges_per_bucket[bucket_offset + dst_bucket], 1);
+    }
   }
 }
 template <typename PairT, typename SizeT, typename ContextT>
```
porumbes, line 54: Looks like the same code in `map/insert.cuh`

gunrock/graph/dynamic_graph/slabhash_graph_base.cuh
```diff
@@ -42,28 +42,30 @@ struct SlabHashGraphBase {
...
+  SlabHashGraphBase(){};
   // todo: add node values
   /**
    * @brief Allocate maximum capacity memory for SlabHash graph.
    *
    * @param[in] max_nodes Maximum number of nodes that the graph can store
    * @param[in] max_buckets Maximum number of buckets that the graph will use
    */
-  SlabHashGraphBase(SizeT max_nodes = 1 << 20, SizeT max_buckets = 1 << 20) {
+  void Allocate(SizeT max_nodes = 1 << 20, SizeT max_buckets = 1 << 20) {
```
porumbes, line 61: Consider replacing the magic values `1 << 20`

gunrock/graph/csr_parallel_iterator.cuh
maawad: I talked about this with @neoblizz and he preferred doing this in a different file. My first version had the iterator code inside the `struct Csr` and it did look a bit simpler when you construct the iterator but `csr.cuh` has a lot of code inside it already. I like a different file as well, but either way is fine with me.

gunrock/graph/graph_iterator.cuh
```diff
@@ -0,0 +1,50 @@
...
+ * structures
+ *
+ * @tparam VertexT Vertex identifier type.
+ * @tparam SizeT Graph size type.
+ * @tparam ValueT Associated value type.
+ * @tparam GraphFlag graph flag
+ */
+template <typename VertexT, typename SizeT, typename ValueT, GraphFlag FLAG,
+          GraphFlag GraphType = FLAG& TypeMask>
+struct ParallelIterator {
```
porumbes, line 33: Should / could this live in `graph_base.cuh` ?

gunrock/graph/dynamic_graph/dynamic_graph_parallel_iterator.cuh
maawad: Same as above. I personally prefer separating graph files into three parts: 1) the basic class code, 2) the iterator on top of that, 3) the algorithms on top of 1, 2. But I'm open to ideas here as well.

gunrock/graph/dynamic_graph/dynamic_graph_parallel_iterator.cuh
maawad: Yeah, you are right. I should move all the iterator code to `SlabHashGraphMap/Set/Base` files and this file will only contain calls to these functions.

gunrock/graph/dynamic_graph/slabhash_graph_base.cuh
maawad: Once we have an application, I'll add these two numbers as optional command line parameters. But I have to set them somewhere for now.

neoblizz:
PR is gonna break because of submodule changes for docs.

gunrock/graph/dynamic_graph/dynamic_graph_base.cuh
maawad: Fixed ec8326a.

gunrock/graph/csr_parallel_iterator.cuh
porumbes: Let's go with what you have.

gunrock/graph/dynamic_graph/dynamic_graph_parallel_iterator.cuh
porumbes: Same as above -- let's roll with what you have now.

PR #744, opened 2020-05-06
Fix mgpu memory issues from seperate stream

crozhon:
After updating moderngpu from 1.0 -> 2.0 we forgot to make sure that the
mgpu calls were on the same stream as our other kernels. Operators using
mgpu's SortedSearch and Scan functions expected them to be complete
before the next kernel call, but since the mgpu context was on the
default stream it executed in parallel with the other kernels. This
caused the output_offsets array used in the following kernel to be incorrect,
and out of bounds memory accesses would occur.

This fix ensures that the mgpu::standard_context_t is setup to use the
same stream as the other kernels. To do so, we keep a context pointer,
allocate it on the heap when the operator parameters are intialized, and
free it when they are released.

### DIFF
--- a/gunrock/app/enactor_types.cuh
+++ b/gunrock/app/enactor_types.cuh
@@ -15,7 +15,7 @@
 #pragma once
 
 #include <time.h>
-// #include <moderngpu/context.hxx>
+#include <moderngpu/context.hxx>
 #include <gunrock/util/multithreading.cuh>
 #include <gunrock/util/kernel_runtime_stats.cuh>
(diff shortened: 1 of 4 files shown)


crozhon:
Just using Jenkins to test the build and then merging

jowens:
Nice debugging!

crozhon:
Jenkins approves so I'm bringing this in - @achalagarwal this should solve the errors you were encountering on dev. Let me know if you still experience issues at all.

PR #752, opened 2020-05-26
Dynamic graph iterator organized

maawad:
- Separated the dynamic graph iterator and slab hash graph iterator
- Fixed bug in tests (added test graphs freeing)
- Added missing const qualifier in iterators

### DIFF
--- a/gunrock/graph/csr_parallel_iterator.cuh
+++ b/gunrock/graph/csr_parallel_iterator.cuh
@@ -34,7 +34,7 @@ struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_CSR> {
       graph::Csr<VertexT, SizeT, ValueT, FLAG & graph::HAS_CSR_MASK,
                  cudaHostRegisterDefault, (FLAG & graph::HAS_CSR) != 0>;
 
-  __host__ __device__ ParallelIterator(const VertexT v, const CsrT* graph)
+  __host__ __device__ ParallelIterator(const VertexT v, CsrT* graph)
       : v(v), graph(graph) {
     v_offset = graph->row_offsets[v];
   }
@@ -53,7 +53,7 @@ struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_CSR> {
  private:
   const VertexT v;
   SizeT v_offset;
-  const CsrT* graph;
+  CsrT* graph;
 };
 
 }  // namespace graph
--- a/gunrock/graph/csr_parallel_iterator.cuh
+++ b/gunrock/graph/csr_parallel_iterator.cuh
@@ -34,7 +34,7 @@ struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_CSR> {
       graph::Csr<VertexT, SizeT, ValueT, FLAG & graph::HAS_CSR_MASK,
                  cudaHostRegisterDefault, (FLAG & graph::HAS_CSR) != 0>;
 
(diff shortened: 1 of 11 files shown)


gunrock/graph/csr_parallel_iterator.cuh
```diff
@@ -34,7 +34,7 @@ struct ParallelIterator<VertexT, SizeT, ValueT, FLAG, HAS_CSR> {
       graph::Csr<VertexT, SizeT, ValueT, FLAG & graph::HAS_CSR_MASK,
                  cudaHostRegisterDefault, (FLAG & graph::HAS_CSR) != 0>;
 
-  __host__ __device__ ParallelIterator(VertexT v, CsrT* graph)
+  __host__ __device__ ParallelIterator(const VertexT v, const CsrT* graph)
```
porumbes, line 37: Let's consider naming this something like `ConstParallelIterator`

maawad: How does this (https://github.com/gunrock/gunrock/pull/752/commits/40e9032b4dc4bcfca4b8ff887e887bfa7a4317fc) look?

gunrock/graph/dynamic_graph/slabhash_graph_parallel_iterator.cuh
```diff
@@ -0,0 +1,146 @@
...
+      SizeT num_collision_slabs = 0;
+      SizeT bid;
+      const SizeT v_buckets_offset = graph->d_buckets_offset[v];
+      for (bid = 0; bid < num_buckets; ++bid) {
+        SizeT b_edges = graph->d_edges_per_bucket[bid + v_buckets_offset];
+        num_collision_slabs +=
+            (b_edges + graph->edgesPerSlab - 1) / graph->edgesPerSlab;
+        num_collision_slabs =
+            b_edges == 0 ? ++num_collision_slabs : num_collision_slabs;
+        num_collision_slabs--;
```
porumbes, line 99: I find this hard to follow. Consider adding some more comments describing the "what" and/or "why" of parts of this code. I can see the "how" from the code, but seems odd that no matter what happens, the code `num_collision_slabs--` decrement by one. It seems like there's always an off-by-one error? Can we remove line 99 and subtract one on line 96? If we can't do that, it might be useful to know "why" (or "what" problem this avoids).

gunrock/graph/dynamic_graph/slabhash_graph_parallel_iterator.cuh
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return neighborVal;
+  }
+
+ private:
+  const VertexT v;
+  GraphT* graph;
+  bool slab_not_cached;
+  VertexT neighborId;
+  ValueT neighborVal;
```
porumbes, line 136: I recommend making it obvious that private data members are indeed private. Consider adding an `_` to private member variables (see below). I don't know if we have a "coding guidelines" or "naming conventions" for Gunrock (if we do, let's follow it, if not consider this the start :)). Also, let's be consistent with our naming (e.g., don't mix `camelCase` with `not_camel_case`).

>  private:
  const VertexT v_;
  GraphT* graph_;
  bool slab_not_cached_;
  VertexT neighbor_id_;
  ValueT neighbor_val_;

maawad: Fixed in https://github.com/gunrock/gunrock/pull/752/commits/bc3c5b5d986087863fce0b734fda79e80d38832b.

PR #753, opened 2020-05-27
Adding graph switch for intersect 

maawad:
Changes make sure that the Intersection kernel (which assumes input CSR graph) is only called when the graph is CSR.

### DIFF
--- a/gunrock/oprtr/intersection/kernel.cuh
+++ b/gunrock/oprtr/intersection/kernel.cuh
@@ -50,7 +50,7 @@ template <OprtrFlag FLAG, typename InKeyT, typename OutKeyT, typename SizeT,
               true
 #else
               false
-#endif 
+#endif
           >
 struct Dispatch {
 };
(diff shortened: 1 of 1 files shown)


gunrock/oprtr/intersection/kernel.cuh
```diff
@@ -164,38 +164,69 @@ __launch_bounds__(Dispatch<FLAG, InKeyT, OutKeyT, SizeT, ValueT, VertexT,
...
+                                ParametersT &parameters, InterOpT inter_op) {
+    typedef typename GraphT ::CsrT CsrT;
+    typedef typename FrontierInT ::ValueT InKeyT;
+    typedef typename FrontierOutT::ValueT OutKeyT;
+    typedef typename ParametersT ::SizeT SizeT;
+    typedef typename ParametersT ::ValueT ValueT;
+    typedef typename ParametersT ::VertexT VertexT;
+    typedef typename ParametersT ::LabelT LabelT;
+    typedef typename Dispatch<FLAG, InKeyT, OutKeyT, SizeT, ValueT, VertexT,
+                              InterOpT, true>::KernelPolicyT KernelPolicyT;
```
porumbes, line 197: Consider `using` instead of `typedef` for readability.

>using CsrT = typename GraphT::CsrT
using InKeyT = typename FrontierInT::ValueT
.... etc.
using KernelPolicyT = typename Dispatch<FLAG, InKeyT, O.......

maawad: Discussed with @porumbes and we will leave this unchanged for now for consistency with other gunorck files.

PR #756, opened 2020-06-03
SetDevice after spawning OpenMP threads

crozhon:
By default all newly created threads are associated with the default
device (see reference). This means newly spawned threads must switch
their device if they are not supposed to use the default device.
When RemoveEdges did not set the device as it should, the graph was
allocated to the wrong GPU. Then subsequent kernels would not find it on
the device they ran on.

In the future I recommend attaching some information about the allocated
device to the underlying Array1D. Without this information, it's
impossible to tell whether a given device pointer is invalid or valid on
a different GPU.


crozhon:
PR for build and test.

crozhon:
Have confirmed this fixes the issue locally and on Daisy. Builds and passes Jenkins (except codecov).

PR #767, opened 2020-06-08
Fix Release issue for multiple parameter switches

crozhon:
I added single graph release to the problem_base.cuh code a bit back.
But that was a mistake. When num_gpus == 1, we use SetPointer to create
our sub_graphs array, rather than allocating individual components. So
when num_gpus == 1, no Release should be happening in the problem. This
caused the graph to be invalid during subsequent runs when multiple
parameters were set.


crozhon:
Build and test

Issue #401, opened 2018-10-04
Googletests


neoblizz:
**[difficulty]** description of category
- [ ] individual test

**[medium]** Core textbook primitives in `gtest`:
- [ ] pagerank
- [ ] maxflow
- [ ] minimum spanning tree
- [ ] triangle counting

**[hard]** Correctness checking for hive apps using `gtest`:
- [ ] sparse fused lasso
- [ ] graph search
- [ ] community detection louvain
- [ ] geolocation
- [ ] scan statistics
- [ ] vertex nomination
- [ ] graph projection
- [ ] application classification

**[easy]** Testing command-line parameters:
- [ ] write a test that just calls a simple `help` command line, listing every available command-line for a primitive or a set of primitive, this drastically improves code coverage.
- [ ] make sure that the given parameters are usable and relevant to the application.
- [ ] `LB`, `LB_CULL`, (different load balancing strategies), note that this is the core output (so, correct or inccorrect), should be easy to detect.
- [ ] Try different sources for `sssp` (for example).
- ... (long list, we have a lot of command line parameters).

**[hard]** Operator level tests:
- [ ] write an simple application that does `advance` only.
- [ ] Advance with a counter or atomics test.
- [ ] write an simple application that does `filter` only.
- [ ] filter with a `boolean`, `atomics` or simple compute.
- [ ] write an simple application that does `ForAll()` and `ForEach()`.

**[medium]** CPU and testing routines:
- [ ] write a test that invokes validation code for all available applications.
- [ ] may require boost if CPU implementation is boost dependent.

**[medium]** `graphio` testing:
- [ ] test different cases of graph types:
- [ ] test invalid/valid market formats
- [ ] test invalid/valid labels formats
- [ ] test random edge weights generator
- [ ] test static (1) edge weights
- [ ] test invalid/valid `mtx` file names

PR #783, opened 2020-07-23
Re #401 : Add tests to call --help for existing primitives

yuqli:
Hope it's not too intruding but I read on this issue list : https://github.com/gunrock/gunrock/issues/401 that adding simple tests to call "--help" for primitives could be helpful to increase code coverage. Added them in CMake.


neoblizz:
@yuqli we welcome pull requests, so thank you! I believe I solved this and didn't mark it as done. Do you think your PR does something different than what I have here: 

https://github.com/gunrock/gunrock/blob/master/cmake/SetSubProject.cmake#L53-L55

Happy to accept more PRs if you'd like to take a shot at things. Feel free to @ me if you have any questions.

yuqli:
@neoblizz Yes it's the same thing but your way is better. I was just browsing through that list and trying to get familiar with the code base. Thanks for the timely reply! 

Also cool library :)

PR #798, opened 2020-09-30
Local Clustering Coefficient 

achalagarwal:
Implementing Local Clustering Coefficient as outlined in  https://ldbc.github.io/ldbc_graphalytics_docs/graphalytics_spec.pdf#page=39

The LCC required in graphalytics is a different community detection algorithm than the Louvain Clustering Coefficient that we have implemented in Gunrock.

This pull request implements the Local Clustering Coefficient which naturally derives from the triangle counting algorithm that we have.


achalagarwal:
Is this still safe to merge? 😬

crozhon:
Most likely. If not we can fix it in dev.

PR #835, opened 2021-01-13
Add multiple C++ interface for multiple algorithms

cameronshinn:
Adding algorithms to C++ interface by doing the following for each:
1. Change `*_app.cu` -> `*_app.cuh`
2. Move `num_runs` argument to the end, but not after `memspace` arguments
3. Remove default arguments from `*_app.cuh`
4. Add algorithm to `gunrock.hpp`


cameronshinn:
Still need to add some tests that use the CPP interface, but the current tests haven't been breaking after modifying the function arguments.

neoblizz:
Out of date now.

PR #842, opened 2021-01-25
k-Core Vertex Decomposition App

angeil:
Computes the k-core value for all vertices in the input graph. A k-core is a maximal subgraph in which all vertices have degree >= k. A vertex’s k-core value is the largest value of k for which it is a member of a k-core. 

How it works:
To find the k-cores, start with the full graph and set k=2. Then iteratively remove vertices (mark them as deleted) with degree <= k, set those vertices’ k-core values to the current value of k, and decrement the degree for all neighbor vertices. Increment k when all vertices with degree <= k have been removed. Repeat the vertex removal process. Continue until no vertices remain.

Possible bug:
Maybe/probably has data race bug(s) that can occasionally cause incorrect results for some vertices on some runs.
Can have results that are lower than CPU version (i.e., GPU-computed k-core = 4 and CPU-computed k-core = 5) OR can have results that are higher than CPU version (i.e., GPU-computed k-core = 5 and CPU-computed k-core = 4). They always seems to be off by only 1.

I can think of a few ways this could happen:
- If the neighbors’ degrees are not properly decremented in the filter operator.
- If the correct vertices are not passed into the filter input frontier by the advance operator.
- If the filter operator does not output the correct vertices to pass back to the advance operator.

When I run cuda-memcheck --tool racecheck, it returns a lot of warnings, generally related to warp_scan.cuh. Example message:
```
========= WARN:(Warp Level Programming) Potential RAW hazard detected at __shared__ 0xf6 in block (105, 0, 0) :
=========     Write Thread (24, 0, 0) at 0x00001190 in /home/afton/Projects/maxClique/gunrock/examples/kcore/../../gunrock/util/scan/warp_scan.cuh:151:_ZN7gunrock5oprtr4CULL6KernelILj1EjjjfjZNS_3app5kcore18kcoreIterationLoopINS4_7EnactorINS4_7ProblemINS3_9TestGraphIjjfLj1808ELj0EEELj0EEELj0ELj0EEEE4CoreEiEUlRKjRjSE_SE_SE_SF_E0_EEvbT2_PKT0_PKT3_SH_T4_PSO_PhPT1_NS_4util15CtaWorkProgressISH_EET5_
=========     Read Thread (25, 0, 0) at 0x000011a8 in /home/afton/Projects/maxClique/gunrock/examples/kcore/../../gunrock/util/scan/warp_scan.cuh:160:_ZN7gunrock5oprtr4CULL6KernelILj1EjjjfjZNS_3app5kcore18kcoreIterationLoopINS4_7EnactorINS4_7ProblemINS3_9TestGraphIjjfLj1808ELj0EEELj0EEELj0ELj0EEEE4CoreEiEUlRKjRjSE_SE_SE_SF_E0_EEvbT2_PKT0_PKT3_SH_T4_PSO_PhPT1_NS_4util15CtaWorkProgressISH_EET5_
=========     Current Value : 0
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/usr/lib/x86_64-linux-gnu/libcuda.so.1 (cuLaunchKernel + 0x2cd) [0x23d14d]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x17167b]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x18edde]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0xea15b]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0xead24]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0xee1d9]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0xef544]
=========     Host Frame:/usr/lib/x86_64-linux-gnu/libgomp.so.1 (GOMP_parallel + 0x42) [0x11232]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x131b96]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x134a3c]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x14eacf]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x15000d]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x1529ee]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x7994]
=========     Host Frame:/lib/x86_64-linux-gnu/libc.so.6 (__libc_start_main + 0xf0) [0x20840]
=========     Host Frame:./bin/test_kcore_9.0_x86_64 [0x7ae9]
```


gunrock/app/hello/hello_test.cuh
```diff
@@ -61,10 +61,10 @@ double CPU_Reference(const GraphT &graph,
  * @brief Validation of hello results
  * @tparam     GraphT        Type of the graph
  * @tparam     ValueT        Type of the values
- * @param[in]  parameters    Excution parameters
+ * @param[in]  parameters    Execution parameters
```
crozhon, line 64: thanks for fixing these

gunrock/app/kcore/kcore_enactor.cuh
```diff
@@ -0,0 +1,578 @@
...
+
+#include <gunrock/app/enactor_base.cuh>
+#include <gunrock/app/enactor_iteration.cuh>
+#include <gunrock/app/enactor_loop.cuh>
+#include <gunrock/oprtr/oprtr.cuh>
+
+#include <gunrock/app/kcore/kcore_problem.cuh>
+
+//#define PRINT_DEBUG(FLAG, ID, K) (FLAG && (ID == 3366) && (K <= 129))
+#define PRINT_DEBUG(FLAG, ID, K) false
```
crozhon, line 25: Should this be hard-coded to false?

gunrock/app/kcore/kcore_problem.cuh
```diff
@@ -0,0 +1,382 @@
...
+      GUARD_CU(k_cores.Allocate(sub_graph.nodes, target));
+      GUARD_CU(initial_frontier.Allocate(sub_graph.nodes, target));
+      GUARD_CU(empty_flag.Allocate(1, target));
+      GUARD_CU(delete_bitmap.Allocate(((sub_graph.nodes / 32) + 1), target));
+      GUARD_CU(to_be_deleted_bitmap.Allocate(((sub_graph.nodes / 32) + 1), target));
+      // </TODO>
+
+      GUARD_CU2(cudaDeviceSynchronize(), "cudaDeviceSynchronize failed.");
+
+      if (target & util::DEVICE) {
```
crozhon, line 145: What if the graph is already allocated on the GPU when Problem::Init is called, but `target == util::DEVICE`? You can see it handled in hits_problem.cuh: https://github.com/gunrock/gunrock/blob/a7fc6948f397912ca0c8f1a8ccf27d1e9677f98f/gunrock/app/hits/hits_problem.cuh#L174

examples/kcore/test_kcore.cu
```diff
@@ -0,0 +1,142 @@
...
+   */
+  template <typename VertexT,  // Use int as the vertex identifier
+            typename SizeT,    // Use int as the graph size type
+            typename ValueT>   // Use int as the value type
+  cudaError_t
+  operator()(util::Parameters &parameters, VertexT v, SizeT s, ValueT val) {
+    // CLI parameters
+    bool quick = parameters.Get<bool>("quick");
+    bool quiet = parameters.Get<bool>("quiet");
+    parameters.Set("undirected", true);
```
crozhon, line 47: So you require graphs to be undirected?

gunrock/app/kcore/kcore_app.cu
```diff
@@ -0,0 +1,321 @@
...
+  bool quiet_mode = parameters.Get<bool>("quiet");
+  int num_runs = parameters.Get<int>("num-runs");
+  std::string validation = parameters.Get<std::string>("validation");
+  util::Info info("kcore", parameters, graph);
+
+  util::CpuTimer cpu_timer, total_timer;
+  cpu_timer.Start();
+  total_timer.Start();
+
+  // <TODO> get problem specific inputs, e.g.:
```
crozhon, line 78: Are these TODOs necessary?

gunrock/app/kcore/kcore_app.cu
```diff
@@ -0,0 +1,321 @@
...
+    cpu_timer.Start();
+    GUARD_CU(enactor.Enact(
+        // <TODO> problem specific data if necessary:
+        // srcs[run_num % srcs.size()]
+        // </TODO>
+        ));
+    cpu_timer.Stop();
+    info.CollectSingleRun(cpu_timer.ElapsedMillis());
+
+    /*util::PrintMsg(
```
crozhon, line 121: Remove commented line

gunrock/app/kcore/kcore_app.cu
```diff
@@ -0,0 +1,321 @@
...
+    SizeT num_errors = Validate_Results(parameters, graph, h_k_cores, ref_k_cores, false);
+  }
+
+  // compute running statistics
+  // <TODO> change NULL to problem specific per-vertex visited marker, e.g.
+  // h_distances
+  info.ComputeTraversalStats(enactor, (VertexT *)NULL);
+// Display_Memory_Usage(problem);
+#ifdef ENABLE_PERFORMANCE_PROFILING
+  // Display_Performance_Profiling(&enactor);
```
crozhon, line 153: Should this be commented?

gunrock/app/kcore/kcore_app.cu
```diff
@@ -0,0 +1,321 @@
...
+
+  info.Finalize(cpu_timer.ElapsedMillis(), total_timer.ElapsedMillis());
+  return retval;
+}
+
+}  // namespace kcore
+}  // namespace app
+}  // namespace gunrock
+
+// ===========================================================================================
```
crozhon, line 175: You can delete everything after here.

gunrock/app/kcore/kcore_enactor.cuh
```diff
@@ -0,0 +1,578 @@
...
+
+    //Fill frontier with all vertices in graph for next iteration
+    SizeT num_nodes = graph.nodes;
+    frontier.queue_length = num_nodes;
+    GUARD_CU(frontier.V_Q()->ForEach(
+      initial_frontier,
+      [] __host__ __device__(VertexT & v, VertexT & i) { v = i; 
+      }, num_nodes, util::DEVICE, 0));
+
+    while(true) {
```
crozhon, line 211: Why do you loop from within Core here? Core should get repeatedly called until StopCondition() == true, so your loop would be repeated too.

gunrock/app/kcore/kcore_enactor.cuh
```diff
@@ -0,0 +1,578 @@
...
+    return retval;
+  }
+
+  /**
+   * @brief Enacts a k-core computation on the specified graph.
+...
+   * \return cudaError_t error message(s), if any
+   */
+  cudaError_t Enact(
+      // <TODO> problem specific data if necessary, eg
```
crozhon, line 560: Remove unnecessary TODOs

gunrock/app/kcore/kcore_problem.cuh
```diff
@@ -0,0 +1,382 @@
...
+ * @brief Specifying parameters for k-core Problem
+ * @param  parameters  The util::Parameter<...> structure holding all parameter
+ * info \return cudaError_t error message(s), if any
+ */
+cudaError_t UseParameters_problem(util::Parameters &parameters) {
+  cudaError_t retval = cudaSuccess;
+
+  GUARD_CU(gunrock::app::UseParameters_problem(parameters));
+
+  // <TODO> Add problem specific command-line parameter usages here, e.g.:
```
crozhon, line 33: Remove unnecessary TODOs

gunrock/app/kcore/kcore_problem.cuh
```diff
@@ -0,0 +1,382 @@
...
+
+    if (this->num_gpus == 1) {
+      auto &data_slice = data_slices[0][0];
+
+      // Set device
+      if (target == util::DEVICE) {
+        GUARD_CU(util::SetDevice(this->gpu_idx[0]));
+
+        // <TODO> extract the results from single GPU, e.g.:
+        GUARD_CU(data_slice.k_cores.SetPointer(h_k_cores, nodes, util::HOST));
```
crozhon, line 264: h_k_cores is not necessarily a host-side pointer. It could be device side. See hits_problem.cuh::Extract()

gunrock/app/kcore/kcore_problem.cuh
```diff
@@ -0,0 +1,382 @@
...
+            },
+            nodes, util::HOST));
+
+        // </TODO>
+      }
+    } else {  // num_gpus != 1
+
+      // ============ INCOMPLETE TEMPLATE - MULTIGPU ============
+
+      // // TODO: extract the results from multiple GPUs, e.g.:
```
crozhon, line 282: Just delete this commented code.

gunrock/app/kcore/kcore_test.cuh
```diff
@@ -0,0 +1,172 @@
...
+                                        // </TODO>
+                                        bool verbose = true) {
+  typedef typename GraphT::VertexT VertexT;
+  typedef typename GraphT::SizeT SizeT;
+
+  SizeT num_errors = 0;
+  bool quiet = parameters.Get<bool>("quiet");
+  bool quick = parameters.Get<bool>("quick");
+
+  // <TODO> result validation and display
```
crozhon, line 135: TODOs

examples/kcore/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# ------------------------------------------------------------------------
+#  Gunrock: Sub-Project k-Core Vertex Decomposition
+# ------------------------------------------------------------------------
+project(kcore)
+message("-- Project Added: ${PROJECT_NAME}")
+include(${CMAKE_SOURCE_DIR}/cmake/SetSubProject.cmake)
```
crozhon, line 6: Add test here. See example: https://github.com/gunrock/gunrock/blob/master/examples/hits/CMakeLists.txt

crozhon (commented):
Thanks you so much for fixing all those spelling mistakes 

The unnecessary TODOs and commented code should be removed. Could you also handle the case where a graph is already allocated on device? It makes it easier for Cameron and I to add a C++ interface in the future. Also could you add a small test to CMakeLists.txt that uses your built application? Like this for HITS: https://github.com/gunrock/gunrock/blob/master/examples/hits/CMakeLists.txt It just uses one of the tiny graphs in dataset/small. 

After that, looks good to me!

gunrock/app/kcore/kcore_enactor.cuh
angeil: It's just for print debugging when/if I have a reproducible bug, so basically I just set it to off to avoid printing a bunch of stuff. I never removed these because I'm still not sure if I fixed the bug, but I can do that if you think I should.

gunrock/app/kcore/kcore_problem.cuh
angeil: Done, hopefully correctly? https://github.com/gunrock/gunrock/pull/842/commits/8dea4711205218851e6e9ef731c29ff6fdf74091

examples/kcore/test_kcore.cu
angeil: Yes, mainly because that's what I need for my maximum clique implementation. Lol. It's also what I've seen in the few other k-core papers I've looked at. But I think an "in-degree k-core" or "out-degree k-core" are concepts that exist, and it would be nice to add versions that compute those in the future.

gunrock/app/kcore/kcore_app.cu
angeil: I have no idea. It seems like this was commented out in `hello_app.cu`, and that's what I used that as the template for this app.

gunrock/app/kcore/kcore_problem.cuh
angeil: https://github.com/gunrock/gunrock/pull/842/commits/8dea4711205218851e6e9ef731c29ff6fdf74091

examples/kcore/CMakeLists.txt
angeil: Done! https://github.com/gunrock/gunrock/pull/842/commits/0f855c5cbe66df358f2a9e32bf7b3b370ab31a40

gunrock/app/kcore/kcore_enactor.cuh
angeil: I basically want nested while loops:
- The inner loop starts with a frontier of all vertices, and ~removes vertices with degree <= k, decrementing the neighbors' degrees, repeating until there are no more vertices with degree <= k. 
- Then the outer loop checks if there are any vertices remaining in the graph, resets the frontier, and increments k. 

Now that you mention it, I suppose I should be able do this with if statements in the stop condition? Though one problem with that could be a potential slow down, because it appears that the stop condition gets called about 5 times for each iteration of Core.

examples/kcore/test_kcore.cu
angeil: Confirmed that directional k-core decomposition does exist: https://rdrr.io/cran/igraph/man/coreness.html

PR #845, opened 2021-01-28
Fix RAW and WAR hazards found using cuda-memcheck

crozhon:
`__syncthreads` was used to fix the hazards, but it might be unnecessary
to go that far.

Do you know more about these scan functions @neoblizz? Does CUB have this same function? Should we just use that?

### DIFF
--- a/gunrock/util/scan/cooperative_scan.cuh
+++ b/gunrock/util/scan/cooperative_scan.cuh
@@ -287,6 +287,8 @@ struct CooperativeGridScan<RakingDetails, NullType> {
       T exclusive_partial = WarpScan<RakingDetails::LOG_RAKING_THREADS>::Invoke(
           inclusive_partial, warpscan_total, raking_details.warpscan, scan_op);
 
+      __syncthreads();
+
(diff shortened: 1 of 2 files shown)


jowens:
@crozhon can you do a quick perf test to see if this affects runtime?

crozhon:
I don't know if my solution is working yet, so I can't run perf tests. I'm getting major memory errors when I run anything besides ak2010, so this needs further work first.

neoblizz:
> Do you know more about these scan functions @neoblizz? Does CUB have this same function? Should we just use that?

The way the code is structured, the warp scan at least can be replaced by CUB's warp scan. But the Block/Tile stuff maybe a bit harder as it plays into the tiling structure we use for the entire work preparation and blocking. On the topic of race conditions, there are places where if a race is reported in our case, the code is still functionally correct, this seems weird to me as it is showing up in scan now which is part of many other apps (so, `__syncthreads` maybe unnecessary).

crozhon:
Switched to draft. Have to fix .

PR #846, opened 2021-01-28
Fix graph memory allocation and undirected insertion batches

maawad:
- Moved the hash table base slabs allocation to the allocate function (where we know how many base slabs we need)
- Added an option for batched edge insertion functions to avoid doubling edges when the batch is already undirected


porumbes (approved):
🚢 it!

PR #850, opened 2021-02-05
debugging rw v2

bkj:


porumbes (approved):
Comment from @bkj: prior to the fix, indices passed into ForAll operator were local to GPUs.  Fix makes them global to avoid some downstream bugs.

PR #856, opened 2021-02-20
multigpu SAGE, SS, 

crozhon:
Add SAGE multigpu forall.

Collect other changes here as a draft


crozhon:
Replaced ForAlls with mgpu_ForAll. Seems to pass validation

```
CUDA_VISIBLE_DEVICES=0,1 ./bin/sage market /data/gunrock/hive_datasets/mario-2TB/pokec/pokec.mtx --undirected
--------------------------
Run 0 elapsed: 823.848963 ms, #iterations = 25
Embedding validation: PASS
[sage] finished.
 avg. elapsed: 823.848963 ms
 iterations: 25
 min. elapsed: 823.848963 ms
 max. elapsed: 823.848963 ms
 load time: 2135.9 ms
 preprocess time: 3604.080000 ms
 postprocess time: 745.880127 ms
 total time: 5223.563910 ms
```

Going to look at TC now and use this PR to collect the work.

porumbes:
@crozhon can you make sure to add the `graph::GRAPH_UNIFIED` flag in the `test_myapp.cu` files as well. Something like:

`typedef typename app::TestGraph<VertexT, SizeT, ValueT,
                                                             graph::HAS_NODE_VALUES | graph::HAS_CSR | graph::GRAPH_UNIFIED> GraphT;`

crozhon:
Thanks forgot that one. Might explain why I'm having a few memory issues with TC too.

porumbes (approved):

gunrock/oprtr/intersection/kernel.cuh
```diff
@@ -179,21 +178,65 @@ cudaError_t Launch(const GraphT &graph, const FrontierInT *frontier_in,
...
+            rows_offset,
+            cols_offset,
+            frontier_in->GetPointer(util::DEVICE),
+            cols_offset,
+            frontier_out->GetPointer(util::DEVICE),
+            edges_per_gpu,
+            stride,
+            inter_op);
+    }));
+  }
```
porumbes, line 220: I'm not sure about parts of this selection and this bit ` threads.push_back(std::thread( [&, context]()` . The reference in the lambda capture could work, but makes me think a number of the variables could change before the kernel is actually called.

I recommend using a range-based for loop over the `mgpu_context.contexts`and  moving the `rows_offsets` and `cols_offsets` inside of the `threads` lambda function. You can then use the `context.device_id` to compute your offsets (the `device_id` is guaranteed to match / have the same range of values as your `i` in the currently implemented code).

crozhon: Thanks this is helpful.

gunrock/oprtr/intersection/kernel.cuh
```diff
@@ -182,40 +182,38 @@ cudaError_t Launch(gunrock::util::MultiGpuContext mgpu_context,
...
-    auto cols_offset = edges_per_gpu * i + graph.CsrT::column_indices.GetPointer(util::DEVICE);
-
-    printf("Rows offset %p\n", rows_offset);
-    printf("Cols offset %p\n", cols_offset);
+  for (auto& context : mgpu_context.contexts) {
 
     threads.push_back(std::thread( [&, context]() {
-        
+        SizeT edges_per_gpu = gunrock::util::ceil_divide(input_length, mgpu_context.getGpuCount());
+
```
porumbes, line 196: I believe at the moment this only needs to happen once and can be moved outside of the range-based for loop. 

That being said, there could be a future where we're smarter about how many `edges_per_gpu` we dole out (doesn't necessarily have to be the same number for every GPU and could be based on some sort of "load" information). Let's explore that when it becomes an issue.

crozhon:
@cameronshinn Can you pull this and try uncommenting the CUB calls for Louvain? I commented them because I was having some difficulty making them compile. The UNIFIED flag on Array1D changes the type signature, so you'll have to modify cubSortPairs, and I couldn't figure it out.

gunrock/app/app.cu
```diff
@@ -0,0 +1,15 @@
...
+template <typename... Ts>
+auto instantiate() {
+    static auto func = std::tuple_cat(std::make_tuple(
+    	t_sm<Ts,Ts>
+        )...);
+
+    return &func;
+}
+
+template auto instantiate<int>();
```
porumbes, line 15: @crozhon can you help me understand what this is for?

crozhon: Mistake - fixed https://github.com/gunrock/gunrock/commit/b37a2eca2d361c4aa9688da18f798a9ec2aeca9d

PR #858, opened 2021-02-24
cudaMallocManaged, Hints, Read-Only (replaces graphs)

neoblizz:
@porumbes is there a better already existing way you're doing the following?
```cpp
int num_gpus = -1;
cudaGetDeviceCount(&num_gpus);
```

For now, it can just be another hack in `Array1D` to get this working.

### DIFF
--- a/gunrock/oprtr/1D_oprtr/for_all.cuh
+++ b/gunrock/oprtr/1D_oprtr/for_all.cuh
@@ -22,8 +22,8 @@
 namespace gunrock {
 namespace oprtr {
 
-#define FORALL_BLOCKSIZE 256
-#define FORALL_GRIDSIZE 256
+#define FORALL_BLOCKSIZE 1024
+#define FORALL_GRIDSIZE 160
 /*template <
     typename T,
     typename SizeT,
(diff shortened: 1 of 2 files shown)


gunrock/oprtr/1D_oprtr/for_all.cuh
```diff
@@ -255,7 +260,7 @@ cudaError_t mgpu_ForAll(const util::MultiGpuContext& mgpuContext,
 
   // launch ForAll kernel for each gpu (on its own thread)
   std::vector<std::thread> threads;
-  threads.reserve(mgpuContext.getGpuCount());
+  // threads.reserve(mgpuContext.getGpuCount());
```
porumbes, line 263: This is fine to get rid off, but generally nice to reserve space for exactly the number of items needed (if you happen to know ahead of time). Not an issue for the number of threads we're working with, but for a larger number of items the allocator will end up allocating 50-60% more memory than the vector needs (to avoid reallocating and copying).

porumbes (approved):

gunrock/oprtr/1D_oprtr/for_all.cuh
neoblizz: Oops, forgot to uncomment that. I was testing `resize` vs. `reserve`, `reserve` was faster.

PR #860, opened 2021-02-25
Fix louvain to build with multi-GPU

cameronshinn:
@crozhon had the right idea of adding a `ArrayType` template parameter to the templated functions with `util::Array1D` arguments. I extended that to all the other commented out calls that were having this issue.

The other problem I ran into was a linking issue with the [ceil_divide](https://github.com/gunrock/gunrock/compare/multigpu...cameronshinn:cameron-multigpu?expand=1#diff-4969aa931d4336e0aba64479d2dc37fb8abab5af5d2262672d45aa99804431bbL230-L231) function. It was being included in multiple places which made ld unsure of which one to use (despite them all being the same). Making the function `inline` fixes the problem.

https://stackoverflow.com/a/34618587

Note that these changes now make it that so louvain *builds*, but has not been run. I was getting errors trying to build in `examples/louvain/` but had it working with the normal CMake way.

### DIFF
--- a/examples/louvain/Makefile
+++ b/examples/louvain/Makefile
@@ -11,8 +11,8 @@
 
 include ../BaseMakefile.mk
 ALGO = louvain
-EXEC = bin/$(ALGO)_main_$(NVCC_VERSION)_$(ARCH_SUFFIX)
-SRC = $(ALGO)_main.cu
+EXEC = bin/test_$(ALGO)_$(NVCC_VERSION)_$(ARCH_SUFFIX)
+SRC = test_$(ALGO).cu
 
 test: $(EXEC)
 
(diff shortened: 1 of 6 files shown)


porumbes (approved):
Looks good sans a number of end of line changes that aren't really changes (not a huge deal for this commit, but it makes tracking real changes harder). I'll merge the changes and we can figure out your settings before the next PR (likely git line ending configuration).

cameronshinn:
> Looks good sans a number of end of line changes that aren't really changes (not a huge deal for this commit, but it makes tracking real changes harder). I'll merge the changes and we can figure out your settings before the next PR (likely git line ending configuration).

Oops that must have been my editor automatically cutting trailing whitespace.

jowens:
> Oops that must have been my editor automatically cutting trailing whitespace.

As your editor should do!

johari:
You may want to consider integrating a code formatter to the project as well. It made my life significantly easier while I was hacking on Fuchsia (everything before commit went through a `fx format` which I think uses `clang-format` for C++), and it certainly handles trailing whitespaces, and much more!

It was godsend whenever I had the pleasure to use one in a project, and it entirely liberates me from worrying about formatting issues. Definitely suggest.

PR #900, opened 2021-11-07
Use external CUB API instead of internal ones

### LINKED ISSUE #857: Using internal CUB APIs
We currently use some of the internal APIs for CUB. These aren't documented or publicized, so we shouldn't be using them. Ex: DispatchRadixSort in sort_device.cuh. Our behavior is apparently bugged when using these functions in updated versions of CUDA/CUB. 

There shouldn't be any additional functionality we need from the internal APIs, so we just need to use the proper external API. 

This could be a good issue for someone new or learning. They would have to update the old calls to CUB to use the officially documented versions. Then test their changes for correctness. Requires refactoring and reading documentation. But there may be dragons.

AKKamath:
Internal CUB APIs in sort, select, and reduce are changed to external ones.
Fixes #857.


crozhon:
Going to bring this into dev first

PR #902, opened 2021-11-12
Use external CUB API instead of internal ones

AKKamath:
Fixed some parameters that were incorrectly passed in #900


crozhon:
Resolved your commit - in the future you want to PR from a separate topic branch for your changes. So `git checkout -b external-cub`.

PR #920, opened 2022-09-21
Add performance analysis

annielytical:
I have created a new branch since the dev branch here has a different history.

Performance analysis added for:
* BC
* BFS
* PR
* SSSP

for comparison to the original Gunrock. 

BC uses the frontier to track the edges / nodes visited, PR uses the graph's number of edges (the number visited is constant in each iteration), and BFS and SSSP increment values within the lambda functions. Each method possible for each algorithm was tested to confirm results and choose the method with the best performance. 

CLA parsing was added with options to turn performance analysis on, set the JSON output location, etc. The default naming conventions match the original Gunrock. 

performance.hxx was added to aggregate and output performance analysis information.

An example call: ./bin/bfs -m ../datasets/chesapeake/chesapeake.mtx --performance


examples/algorithms/bc/bc.cu
```diff
@@ -1,32 +1,100 @@
...
+
+    if (result.count("json_dir") == 1) {
+      json_dir = result["json_dir"].as<std::string>();
+    }
+
+    if (result.count("json_file") == 1) {
+      json_dir = result["json_file"].as<std::string>();
+    }
   }
+};
```
neoblizz, line 72: I noticed that this bit of code is _almost_ exactly the same in bfs, pr and sssp .cu files. Is there a way you could allow `parameters_t` constructor to take the name of the algorithm, unify all of these under `include/gunrock/io/parameters.hxx`, and simply include that header file in all 4 examples? This should reduce the code significantly and promote reuse.

Also, feel free to assign it a namespace, like:
```cpp
namespace gunrock { // main namespace
namespace io {  // input/output main namespace
namespace cli { // maybe?
```

include/gunrock/algorithms/bc.hxx
```diff
@@ -115,6 +125,11 @@ struct enactor_t : gunrock::enactor_t<problem_t> {
 
     auto bc_values = P->result.bc_values;
 
+    auto edges_visited = P->result.edges_visited;
+    auto search_depth = P->result.search_depth;
+
+    auto performance = P->param.performance;
```
neoblizz, line 131: This might be a tedious change (sorry!) but I prefer `collect_metrics` or something like that instead of `performance`.

include/gunrock/algorithms/bc.hxx
```diff
@@ -177,10 +197,15 @@ struct enactor_t : gunrock::enactor_t<problem_t> {
         operators::advance::execute<operators::load_balance_t::merge_path,
                                     operators::advance_direction_t::forward,
                                     operators::advance_io_type_t::vertices,
-                                    operators::advance_io_type_t::none>(
+                                    operators::advance_io_type_t::vertices>(
```
neoblizz, line 200: This is a significant change in terms of the algorithm and should not be required in my opinion, Let's discuss this in detail on slack.

include/gunrock/algorithms/bfs.hxx
```diff
@@ -122,9 +168,15 @@ struct enactor_t : gunrock::enactor_t<problem_t> {
...
-
+    // Track performance stats if performance evaluation is turned on
+    if (performance) {
+      operators::advance::execute<operators::load_balance_t::block_mapped>(
+          G, E, search_with_performance, context);
+      *search_depth = iteration;
+    } else {
+      operators::advance::execute<operators::load_balance_t::block_mapped>(
+          G, E, search, context);
+    }
```
neoblizz, line 179: I prefer a similar solution for betweenness centrality, where you can have an if-else block, and only allow the output frontier for the case where you need it profiled (collecting metrics).

include/gunrock/algorithms/bc.hxx
neoblizz: Also see my comment in `bfs`.

include/gunrock/util/performance.hxx
```diff
@@ -0,0 +1,183 @@
...
+    sysinfo["sysname"] = sysname();
+    sysinfo["release"] = release();
+    sysinfo["version"] = version();
+    sysinfo["machine"] = machine();
+    sysinfo["nodename"] = nodename();
+    jsn->push_back(nlohmann::json::object_t::value_type("sysinfo", sysinfo));
+  }
+};
+
+void get_gpu_info(nlohmann::json* jsn) {
```
neoblizz, line 35: Can you potentially use this insead? https://github.com/gunrock/gunrock/blob/main/include/gunrock/cuda/device_properties.hxx

If the functionality doesn't exist (maybe you need a version of driver and its not in there), please move that from this bit of code in that file or relevant `cuda/*` file and just call it here. Let's talk on slack about this as well.

include/gunrock/util/performance.hxx
```diff
@@ -0,0 +1,183 @@
...
+#include "nlohmann/json.hpp"
+#include <time.h>
+#include <regex>
+#include <filesystem>
+#include <fstream>
+
+using vertex_t = int;
+using edge_t = int;
+
+class Sysinfo {
```
neoblizz, line 11: Naming convention, `system_info_t`.

include/gunrock/util/performance.hxx
```diff
@@ -0,0 +1,183 @@
+#include <sys/utsname.h>
+#include "nlohmann/json.hpp"
+#include <time.h>
+#include <regex>
+#include <filesystem>
+#include <fstream>
+
```
neoblizz, line 7: Needs namespace:
```cpp
namespace gunrock {
namespace util {
namespace stats {
```

neoblizz (changes requested):
I proposed some changes, let's discuss them in detail on slack if you have any questions.

examples/algorithms/bc/bc.cu
```diff
@@ -1,32 +1,100 @@
...
+    if (result.count("num_runs") == 1) {
+      num_runs = result["num_runs"].as<int>();
+    }
+
+    if (result.count("json_dir") == 1) {
+      json_dir = result["json_dir"].as<std::string>();
+    }
+
+    if (result.count("json_file") == 1) {
+      json_dir = result["json_file"].as<std::string>();
```
neoblizz, line 69: Shouldn't this be `json_file = ...`

neoblizz:
Thank you @annielytical, this is great.

PR #923, opened 2022-10-06
Performance Analysis Updates

annielytical:
JSON output schema updates (add git commit SHA-1, compiler information, tags, etc.)
Allow multiple / random starting sources for BC, BFS, SSSP
Collect runtimes with performance analysis turned off 

I still need to add the Windows system info code, so feel free to ignore this until then. I just wanted to submit now in case we want to start runs on the cluster before then.


examples/algorithms/bc/CMakeLists.txt
```diff
@@ -18,4 +18,11 @@ set_target_properties(${APPLICATION_NAME}
...
-# end /* Add CUDA executables */
\ No newline at end of file
+# end /* Add CUDA executables */
+
+# Add GIT SHA-1 definition
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/externals/cmake_modules-src/")
+include(GetGitRevisionDescription)
+get_git_head_revision(GIT_REFSPEC GIT_SHA1)
+
+set_property(SOURCE bc.cu PROPERTY COMPILE_DEFINITIONS GIT_SHA1="${GIT_SHA1}")
```
neoblizz, line 28: The main question/comment I have is this is being repeated for every algorithm, why does this need to be in the individual cmake files and not instead in the main cmake file attached to the gunrock project? For example: https://github.com/gunrock/gunrock/blob/7e098254aec01a28f28a25b3401fc66a37243238/CMakeLists.txt#L87-L94

neoblizz (changes requested):

annielytical:
Latest commit makes change requested above.

PR #1036, opened 2023-03-15
Gunrock with HIP/ROCm Support

atharvag1:
Development branch for the HIP port of gunrock/develop.


neoblizz:
@atharvag1 thank you Atharva! Do you have the known issues documented that you can share here?

atharvag1:
@neoblizz 

Here's a brief summary of the known issues and list of unimplemented portions

1) SPMV and BC: ModernGPU-dependent. I have not ported the apps with ModernGPU dependencies. The reason I did not port modernGPU is because it uses some warp-level primitives that do not have a HIP equivalent as far as I know.

2) The port (https://github.com/atharvag1/gunrock) is functional on both AMD and NVIDIA targets but to build it, you need to load HIP (and CUDA, for NVIDIA targets) using ```module load hip (and/or cuda)``` and you also need rocthrust and hipcub installed in ```/opt/rocm/```
It will not compile for the systems where you cannot load HIP using module commands and if rocthrust and hipcub are not installed


Other than the two issued I mentioned. I was able to verify functional correctness of all other apps.
Feel free to let me know if any changes/updates are required to my branch

PR #1046, opened 2023-04-17
Update Graph Builder API

annielytical:
This PR updates the graph builder API to take in CSR / COO  / CSC struct(s) instead of pointers. This allows the graph to use views for these different formats simultaneously. Graph properties are also now supported. 

Unrelated issues noted during my testing:
- mst: something on develop (prior to my updates) breaks mst (it exits when it shouldn't) but it runs properly on master
- async_bfs: hangs (with and without my updates)
- tc: error on validate (with and without my updates)
- all benchmarks: it looks like using filtered_argv is introducing an encoding issue when --json is used (with and without my updates)


include/gunrock/graph/csr.hxx
```diff
@@ -20,6 +21,11 @@ namespace graph {
 struct empty_csr_t {};
 
 using namespace memory;
+using vertex_t = int;
+using edge_t = int;
+using weight_t = float;
+using csr_t =
+    gunrock::format::csr_t<memory_space_t::device, vertex_t, edge_t, weight_t>;
```
neoblizz, line 28: These should not be hard-coded. vertex, edge or weight could be other types as well.

include/gunrock/graph/detail/build.hxx
```diff
@@ -8,111 +8,229 @@
...
-                         empty_csc_t>;
-
-  // Enable COO.
-  using coo_v_t =
-      std::conditional_t<has(build_views, view_t::coo),
-                         graph::graph_coo_t<vertex_t, edge_t, weight_t>,
-                         empty_coo_t>;
+  using csr_v_t = graph::graph_csr_t<vertex_t, edge_t, weight_t>;
+  using coo_v_t = empty_coo_t;
+  using csc_v_t = empty_csc_t;
```
neoblizz, line 30: If we're explicitly calling the `builder` with whatever views we need, do we need to set unused types to empty? The graph_t struct removes already them.

In the following line you don't specify them:
```cpp
using graph_type = graph::graph_t<space, vertex_t, edge_t, weight_t, csr_v_t>;
```

What is the idea behind this?

include/gunrock/graph/detail/build.hxx
```diff
@@ -8,111 +8,229 @@
...
 
-  using graph_type = graph::graph_t<space, vertex_t, edge_t, weight_t, csr_v_t,
-                                    csc_v_t, coo_v_t>;
+  using graph_type = graph::graph_t<space, vertex_t, edge_t, weight_t, csr_v_t>;
+  graph_type G(properties);
+
+  G.template set<csr_v_t>(csr.number_of_rows, csr.number_of_nonzeros,
+                          memory::raw_pointer_cast(csr.row_offsets.data()),
+                          memory::raw_pointer_cast(csr.column_indices.data()),
+                          memory::raw_pointer_cast(csr.nonzero_values.data()));
```
neoblizz, line 37: Since we are embracing the vector interface, we can replace set to just take the format csr/csc/coo instead of these raw_pointer_cast and set here. And internally it can do the raw_ptrs and set them (it cannot hold structs unfortunately, because that won't work across GPU boundary). Will reduce the `builder` code a lot.

benchmarks/bc_bench.cu
```diff
@@ -73,19 +73,18 @@ void bc_bench(nvbench::state& state) {
...
+  loader = mm.load(filename);
+  
+  format::csr_t<memory_space_t::device, vertex_t, edge_t, weight_t> csr;
+  csr.from_coo(loader.coo);
+
+  // --
+  // Build graph
+
+  auto G =
+      graph::build::build<memory_space_t::device>(loader.properties, csr);
```
neoblizz, line 87: I will probably try to improve this bit. The `.load()` can return two things:
```cpp
auto [properties, coo] = mm.load(filename);
```

https://stackoverflow.com/questions/321068/returning-multiple-values-from-a-c-function

And the graph could then be built:
```cpp
format::csr_t<memory_space_t::device, vertex_t, edge_t, weight_t> csr;
csr.from_coo(coo);
auto G = graph::build::build<memory_space_t::device>(properties, csr);
```

neoblizz (commented):
Some high-level comments.

benchmarks/bc_bench.cu
annielytical: This is updated.

include/gunrock/graph/detail/build.hxx
annielytical: I wanted to do this originally as well, but didn't because `set` is a `__host__ __device__` function and we need to call `.data()`. I updated this and made it a host function.

include/gunrock/graph/detail/build.hxx
annielytical: This is updated. Not sure what I was thinking there.

include/gunrock/graph/csr.hxx
annielytical: The hard-coding is removed.

annielytical:
Thankfully making these updates _magically_ resolved my segfault, so the code **should** be good to go. :)

neoblizz:
> Thankfully making these updates _magically_ resolved my segfault, so the code **should** be good to go. :)

Haha, this is great!

include/gunrock/graph/detail/build.hxx
neoblizz: Oh, I think we can get away without the __device__, so just a regular function.

neoblizz: I thought I needed it when I was doing virtual inheritance, but it was a mess.

annielytical: Ok I got rid of __host__ too so it is just a regular function.

benchmarks/for.cu
```diff
@@ -13,8 +13,8 @@ void parallel_for(nvbench::state& state) {
   // Build a graph using a sample csr.
   auto csr = io::sample::csr();
   auto G =
-      graph::build::from_csr<memory_space_t::device, graph::view_t::csr>(csr);
-
+      graph::build::build<memory_space_t::device>({}, csr);
```
neoblizz, line 16: We can get rid of the namespace `build` because it seems redundant now.

benchmarks/bc_bench.cu
```diff
@@ -73,19 +73,17 @@ void bc_bench(nvbench::state& state) {
...
-  auto G = graph::build::from_csr<memory_space_t::device,
-                                  graph::view_t::csr>(
-      csr.number_of_rows,               // rows
-      csr.number_of_columns,            // columns
-      csr.number_of_nonzeros,           // nonzeros
-      csr.row_offsets.data().get(),     // row_offsets
-      csr.column_indices.data().get(),  // column_indices
-      csr.nonzero_values.data().get()   // values
-  );
+  auto [properties, coo] = mm.load(filename);
```
neoblizz, line 77: I like this! ❤️

examples/algorithms/hits/hits.cu
```diff
@@ -19,29 +19,31 @@ void test_hits(int num_arguments, char** argument_array) {
   // IO
 
   gunrock::io::cli::parameters_t params(num_arguments, argument_array,
-                                        "Hyperlink-Induced Topic Search");
+                                        "Betweenness Centrality");
```
neoblizz, line 22: Incorrect algorithm name. Probably an artifact of copy-pasta. :)

neoblizz (commented):
This looks great, just some minor comments and I can get it merged! Thank you, Annie!

neoblizz:
Also @annielytical, feel free to update headers of the file with you as the author for the new graph build-system!

benchmarks/for.cu
annielytical: Ok this is done!

PR #1055, opened 2023-10-06
Benchmarking rework

annielytical:
This PR reworks our benchmarking code. Instead of tracking metrics within the algorithm's hxx file, the values are incremented within the operator code itself. This method was started by @neoblizz.

The same four algorithms are currently supported: BC, BFS, PR, and SSSP.

The cmake variable ESSENTIALS_COLLECT_METRICS turns the metrics collection on/off. Because this method uses preprocessing, it should have no impact on performance when disabled. Only the above four algorithms can be built with the option enabled. 

Most metrics must be collected with the option on but runtimes must be collected with the option off. The command-line parameter --export_metrics will export the applicable metrics (either all or just runtimes) depending on ESSENTIALS_COLLECT_METRICS.

**The helper Python script collect_metrics.py can be used to build, run, and combine data from both versions.**


neoblizz:
Hi @annielytical, this is really cool, thank you for putting it together. I will review and merge by this weekend.

examples/algorithms/bc/bc.cu
```diff
@@ -54,51 +53,63 @@ void test_bc(int num_arguments, char** argument_array) {
...
+    thrust::host_vector<int> h_edges_visited = benchmark::____.edges_visited;
+    thrust::host_vector<int> h_vertices_visited =
+        benchmark::____.vertices_visited;
 
-      edges_visited_vect.push_back(edges_visited);
-      search_depth_vect.push_back(search_depth);
-    }
+    edges_visited_vect[i] = h_edges_visited[0];
+    vertices_visited_vect[i] = h_vertices_visited[0];
+    search_depth_vect[i] = benchmark::____.search_depth;
```
neoblizz, line 91: I think this should be transparent as well. If we have INIT_BENCH, DESTROY_BENCH, we can just create another wrapper which moves the data to host: `benchmark::EXTRACT()`? And _that_ returns a well-defined struct that contains the benchmarked metrics. The struct can then be document well to explain what each variable means and how it was obtained (which will be the only interface the user will care about).

- The result of the benchmarks (the struct) can be passed in to the `export_performance_stats` function as well, without needing `empty_vector`, because the struct can have a set default when ESSENTIALS_COLLECT_METRICS was false.
- It will also remove the need for creating std::vectors, etc.

examples/algorithms/bc/bc.cu
```diff
@@ -54,51 +53,63 @@ void test_bc(int num_arguments, char** argument_array) {
...
-  // Log
 
-  std::cout << "Single source : " << source_vect.back() << "\n";
-  print::head(bc_values, 40, "GPU bc values");
-  std::cout << "GPU Elapsed Time : " << run_times[params.num_runs - 1]
-            << " (ms)" << std::endl;
-
-  // --
-  // Run performance evaluation
+#if !(ESSENTIALS_COLLECT_METRICS)
```
neoblizz, line 59: The main change I see is that we don't need this #if, #else, #endif. It can just be a part of the `benchmark::INIT_BENCH();`. When the variable is set, the init does something, otherwise it is just empty.

This way, we can always have `benchmark::INIT_BENCH();` there with no if/else branch and duplicated code.

neoblizz (changes requested):
@annielytical I have added the main change I am requesting. Which should further simplify the code and require less code overall. See what you think.

examples/algorithms/bc/bc.cu
neoblizz:
```cpp
void INIT_BENCH() {
#if ESSENTIALS_COLLECT_METRICS
  thrust::fill(____.edges_visited.begin(), ____.edges_visited.end(), 0);
  thrust::fill(____.vertices_visited.begin(), ____.vertices_visited.end(), 0);

  BXXX.d_edges_visited = ____.edges_visited.data().get();
  BXXX.d_vertices_visited = ____.vertices_visited.data().get();
#endif
}
```

annielytical:
@neoblizz thank you for the review and suggestions! I agree those changes definitely improve and simplify the design. I have pushed an update with these changes.

neoblizz (approved):

PR #1061, opened 2023-12-07
Fix MST bug introduced with par_nosync

annielytical:
The switch to par_nosync introduced a bug in MST where it would exit early. An exit condition was checked asynchronously before the value was updated. This fixes that issue.


annielytical:
@neoblizz I am merging this in since it is a bug fix.

